PR #122, opened 2018-02-13
add group and track calls for segment

### LINKED ISSUE #131: Handle null user ids in segment
[Customer](https://looker.zendesk.com/agent/tickets/134428) looks like they have their Segment integration set up correctly per this [doc](https://docs.looker.com/admin-options/platform/segment) but schedules are failing with the error: Could not send webhook to "https://actions.looker.com/actions/segment_event/execute". {"success":false,"error":"Internal server error."} (Webhook ID: 374ad6d90f73421c381c709af2998e2c) 

disqus

Looks similar to https://github.com/looker/helltool/issues/30765 but not exactly the same. 
 
Also opened issue in helltool: https://github.com/looker/helltool/issues/36296

jdavidson:
closes looker/helltool#36296
closes #131


jdavidson:
@wilg i've changed this enough that I'd love another look - mostly just fixing my incorrect understanding of the api calls

also fixed this bug https://github.com/looker/actions/issues/131

src/actions/segment/segment.ts
```diff
@@ -28,6 +44,61 @@ export class SegmentAction extends Hub.Action {
   supportedVisualizationFormattings = [Hub.ActionVisualizationFormatting.Noapply]
   requiredFields = [{ any_tag: this.allowedTags }]
 
+  taggedFields(fields: any, tags: string[]) {
```
wilg: you added tests for these (good!) but if you didn't need to you could mark these as `protected`

src/actions/segment/segment.ts
```diff
@@ -28,6 +44,61 @@ export class SegmentAction extends Hub.Action {
...
+    return fields.filter((f: any) =>
+      f.tags && f.tags.some((t: string) => tags.indexOf(t) !== -1),
+    )
+  }
+
+  taggedField(fields: any, tags: string[]) {
+    return this.taggedFields(fields, tags)[0]
+  }
+
+  segmentFields(fields: any): SegmentFields {
```
wilg: if we import `lookml_model_explore_field` using the fancy api types importer thing and `import LookmlModelExploreField as Field` we could actually type these with real types instead of `any`

PR #138, opened 2018-05-09
Add a streaming interface for actions, and implement it for S3 and DigitalOcean

wilg:
See the documentation for `stream`, `streamJson` and `streamJsonDetail`.


src/actions/digitalocean/test_digitalocean_object_storage.ts
```diff
@@ -5,26 +5,9 @@ import * as Hub from "../../hub"
 
 import { DigitalOceanObjectStorageAction } from "./digitalocean_object_storage"
```
phillipperalez: Per the title I don't see the digital ocean action - is that in a different PR?

src/hub/action.ts
```diff
@@ -42,6 +43,7 @@ export abstract class Action {
   abstract name: string
   abstract label: string
   abstract description: string
+  usesStreaming = false
```
phillipperalez: neat, a lot easier to understand what the variable means

src/hub/action_request.ts
```diff
@@ -118,6 +133,92 @@ export class ActionRequest {
...
+   * ```
+   *
+   * Streaming generally occurs only if Looker sends the data in a streaming fashion via a push url,
+   * however it will also wrap non-streaming attachment data so that actions only need a single implementation.
+   *
+   * @returns The return value of the `callback` function. This is useful for returning
+   * a promise from the `callback` function.
+   * @param callback A function will be caled with a Node.js `Readable` object.
+   * The readable object represents the streaming data.
+   */
```
phillipperalez: slick implementation of generic streams in the ActionRequest :+1:

src/hub/action_request.ts
```diff
@@ -118,6 +133,92 @@ export class ActionRequest {
...
+  stream<T>(callback: (readable: Readable) => T): T {
+    const url = this.scheduledPlan && this.scheduledPlan.downloadUrl
```
phillipperalez: I have a PR that can and will change the downloadUrl location to the attachment in helltool in waiting, that would be an easy thing to add after I am back next week though - I could just edit this part and make the change if it goes in though

wilg: Haven't we shipped it in the previous place though?

wilg: Will that be a backwards compatible change?

phillipperalez: Ah I was changing the location as a response to this: https://github.com/looker/helltool/pull/37008 . I can just undo that change. It won't be backwards compatible unless I put it in both the scheduledPlan and the attachment

package.json
```diff
@@ -33,9 +33,11 @@
     "@sendgrid/mail": "^6.1.4",
     "@slack/client": "^3.14.0",
     "@types/body-parser": "^1.16.5",
+    "@types/concat-stream": "^1.6.0",
```
jdavidson: looks like this is just used in tests. should it be a dev dependency?

src/hub/action_request.ts
```diff
@@ -118,6 +133,92 @@ export class ActionRequest {
...
+   * be called for each row.
+   *
+   * @returns A promise that will be resolved when streaming is complete.
+   * @param onRow A function that will be called for each streamed row, with the row as the first argument.
+   */
+  async streamJson(onRow: (row: { [fieldName: string]: any }) => void) {
+    return new Promise<void>((resolve) => {
+      this.stream((readable) => {
+        oboe(readable)
+          .node("![*]", (row) => {
```
jdavidson: is it worth calling out what the format structure means?

wilg: i think its complex enough that you have to look at the oboe docs anyway, which should be relatively obvious

Issue #53, opened 2017-09-26
AWS S3 Pull region into the form and out of params


jdavidson:
requires https://github.com/looker/integrations/issues/52 so that buckets can be generated

PR #139, opened 2018-05-10
Remove need to specify region for S3 and DigitalOcean

wilg:
and remove seperate file path form field that was ignored

TODO: Still need to validate this with DigitalOcean, their Spaces API is having issues today.

closes https://github.com/looker/actions/issues/33
closes https://github.com/looker/actions/issues/53


src/actions/amazon/amazon_s3.ts
```diff
@@ -81,23 +75,44 @@ export class AmazonS3Action extends Hub.Action {
...
+    bucketName: string,
+  ): Promise<string> {
+    const s3 = this.amazonS3ClientFromRequest(request)
+    try {
+      // If you're the bucket owner this will work
+      const location = await s3.headBucket({ Bucket: bucketName }).promise()
+      return location.$response.httpResponse.headers["x-amz-bucket-region"]
+    } catch (e) {
+      // Otherwise the error (!?) will include the region
+      return e.region
```
jdavidson: wow. that's rough.

PR #140, opened 2018-05-18
Segment - Streaming

jdavidson:
- [x] fix throwing error through oboe or abort stream on error


src/actions/segment/segment.ts
```diff
@@ -1,16 +1,19 @@
 import * as uuid from "uuid"
 
+import { LookmlModelExploreField as Field } from "../../api_types/lookml_model_explore_field"
 import * as Hub from "../../hub"
+import { Row as JsonDetailRow } from "../../hub/json_detail"
```
wilg: should just be able to use `Hub.JsonDetail` with the existing import

jdavidson: much cleaner! thanks

PR #144, opened 2018-05-24
Marketo - Streaming

jdavidson:


src/actions/marketo/test_marketo.ts
```diff
@@ -8,15 +8,28 @@ import { MarketoAction } from "./marketo"
 const action = new MarketoAction()
 
 describe(`${action.constructor.name} unit tests`, () => {
-  describe("action", () => {
+  describe.only("action", () => {
```
wilg: 😑

src/actions/marketo/marketo.ts
```diff
@@ -43,46 +35,52 @@ export class MarketoAction extends Hub.Action {
...
+        fieldset = Hub.allFields(fields)
+        let hasTagMap
+        for (const field of fieldset) {
+          hasTagMap = field.tags.map((tag: string) => {if (tag.startsWith("marketo:")) {
+            return tag.split("marketo:")[1]
+          }})
+          fieldMap[field.name] = hasTagMap.find((tag: string | undefined) => tag !== undefined)
+        }
+      },
+      onRow: async (row) => {
```
wilg: hmm... I'm not sure we can do `async` here because the streaming code won't await it...

dwmintz:
Any update on this one?

dwmintz:
Please to review @wilg ?

src/actions/marketo/marketo.ts
wilg: Does marketo support a batch API?

jdavidson: Yeah. They'll accept up to 300 records in the lead.createOrUpdate.

What is the right way to handle the async process? We need the createOrUpdate result ids prior to adding the ids to the campaign.

dwmintz:
Yup. http://developers.marketo.com/rest-api/bulk-import/

Best,
-d

src/actions/marketo/marketo.ts
wilg: I'm not sure – using it here will _work_ but since it's not waited on the action will return success before it finishes. We can probably make `streamJsonDetail` optionally allow async callbacks and await on them.

dwmintz: (n.b. it's actually 100 records per call, not 300)

phillipperalez:
implemented in a different way

PR #157, opened 2018-06-13
Several fixes for the Raven error handling

wilg:
See individual commits for details.

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -156,12 +156,16 @@ export default class Server implements Hub.RouteBuilder {
       } catch (e) {
         this.logError(req, res, "Error on request")
         if (typeof(e) === "string") {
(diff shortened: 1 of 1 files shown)


src/server/server.ts
```diff
@@ -28,7 +28,7 @@ export default class Server implements Hub.RouteBuilder {
       Raven.config(process.env.ACTION_HUB_RAVEN_DSN, {
         captureUnhandledRejections: true,
         release: statusJson.git_commit,
-        autoBreadcrumbs: true,
+        autoBreadcrumbs: false,
```
phillipperalez: Is this just a precaution?

wilg: yes, it doesn't provide particularly useful debugging info in our case either

PR #159, opened 2018-06-13
Handle errors during download_url web request

wilg:
Currently, an error occurring during the web request to the `download_url` will throw an exception and terminate the process.

This change turns the `stream` function to return a promise that will be rejected upon a streaming failure and allowing that error to be propagated upwards.

### DIFF
--- a/src/hub/action_request.ts
+++ b/src/hub/action_request.ts
@@ -145,8 +145,9 @@ export class ActionRequest {
    * Streaming generally occurs only if Looker sends the data in a streaming fashion via a push url,
    * however it will also wrap non-streaming attachment data so that actions only need a single implementation.
    *
-   * @returns The return value of the `callback` function. This is useful for returning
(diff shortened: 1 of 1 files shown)


src/hub/action_request.ts
```diff
@@ -144,27 +145,47 @@ export class ActionRequest {
...
+    const returnPromise = callback(stream)
+
+    const url = this.scheduledPlan && this.scheduledPlan.downloadUrl
+
+    const streamPromise = new Promise<void>((resolve, reject) => {
+      if (url) {
+        httpRequest
+          .get(url)
+          .on("error", (err) => {
+            winston.error(`Request stream error: ${err}\n${err.stack}`)
```
jdavidson: do you want to keep this logging in?

wilg: Yes, it's useful to have these errors in the action hub logs.

PR #160, opened 2018-06-14
pass error on failure

jdavidson:


src/server/server.ts
```diff
@@ -101,6 +101,9 @@ export default class Server implements Hub.RouteBuilder {
       const action = await Hub.findAction(req.params.actionId, {lookerVersion: request.lookerVersion})
       if (action.hasExecute) {
         const actionResponse = await action.validateAndExecute(request)
+        if (!actionResponse.success) {
+          res.status(500)
```
wilg: I think a 400 here is better if Looker behaves the same way when it encounters one, and maybe a small comment explaining that this is here because of a Looker bug.

PR #186, opened 2018-08-02
Upgrade node to v10.8.0 for improved memory management

phillipperalez:
Local testing has seen a large improvement in memory management. 

Upgrades Node to v10.8.0. v10 will be LTS around October


wilg:
You will need to change `.nvmrc` also

wilg:
And the `Dockerfile`

src/actions/segment/segment.ts
```diff
@@ -215,7 +215,7 @@ export class SegmentAction extends Hub.Action {
   }
 
   protected segmentClientFromRequest(request: Hub.ActionRequest) {
-    return new segment(request.params.segment_write_key)
+    return new segment(request.params.segment_write_key, {flushAt: 500})
```
wilg: Can we put this in a separate PR for clarity?

wilg: Or at least its own commit

phillipperalez: for sure I'll take that out for this

wilg: or rebase the whole thing so that the node upgrade and the segment flush are separate things.

and i'd still like to make this an environment variable with a default 500

wilg: ok cool yea a seperate PR is best that way this one can be squashed

wilg:
The test failure is strange – maybe the Node upgrade changes something subtle?

PR #187, opened 2018-08-03
Add DataRobot action

pavloko:
DataRobot action allows Looker users to export data to DataRobot.


pavloko:
Hey @wilg! Any chance you'll to glance over this PR?

pavloko:
Thanks for taking a look at this.

I'd be happy to make any changes when needed. Please, let me know if there is anything I can do to help.

I believe our DB team has already received a copy and will be getting back to you shortly.

src/actions/datarobot/datarobot.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  private getRequestOptions(request: Hub.ActionRequest) {
+    const options = {
+      url: `${DR_API_URL}/projects/`,
+      headers: {
+        Authorization: `Token ${request.params.datarobot_api_token}`,
+      },
+    }
+
+    const url = request.scheduledPlan && request.scheduledPlan.downloadUrl
```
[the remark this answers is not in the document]
wilg: Update: I've fixed item 2 on the Looker end, which will make it into 5.24, so I'll update this PR to specify this as the minimum Looker version that this action supports.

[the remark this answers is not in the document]
pavloko: Amazing news:) Excited to get this merged.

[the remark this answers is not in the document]
wilg: Yes, basically we'll add `minimumSupportedLookerVersion = "5.24.0"` to this action, which will make it only show up on or after 5.24. Looker 5.24 and later will properly guarantee that it will only send actions in a supported download setting, so yes you'll be able to rely on Looker only sending you the data via URL. It's worth noting that there is a 5 minute time limit to initiate the request via the download URL.

src/actions/datarobot/datarobot.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (url) {
```
pavloko: @wilg Should I also cleanup this condition then?

src/actions/datarobot/datarobot.ts
```diff
@@ -0,0 +1,120 @@
+import * as Hub from "../../hub"
+
+import * as httpRequest from "request-promise-native"
+
+const DR_API_URL = "https://app.datarobot.com/api/v2"
+
+export class DataRobotAction extends Hub.Action {
+
```
pavloko: Are these two properties I need to add?

```
supported_download_settings = ["url"]
minimumSupportedLookerVersion = "5.24.0"
```

wilg: I still need to enhance the `Hub.Action` api to have an enum instead of the `usesStreaming` boolean, so there's not yet an API for specifying you only want URL. (As mentioned in item 1 of https://github.com/looker/actions/pull/187#discussion_r211436837)

pavloko: Got it;)

I've added `minimumSupportedLookerVersion` just like you mentioned and cleanup the code that asserted `downloadUrl`.

pavloko:
@wilg I've been told the Legal matters are now resolved. I've also excluded Dashboard from `supportedActionTypes` as supporting that we'll take more work, and we decided to leave it for next iteration.

wilg:
Awesome! I should be able to finalize this early next week!

pavloko:
@wilg @erin-looker awesome. Thank you!

PR #192, opened 2018-08-14
Fork Processes for Heavy Streaming Actions

phillipperalez:
Local testing has shown many improvements, however I need this to be thoroughly smoked on product-dev first. Still need to add some tests.


src/actions/actions_process.ts
```diff
@@ -0,0 +1,23 @@
...
+        const version = userAgent.split("LookerOutgoingWebhook/")[1]
+        req.lookerVersion = semver.valid(version)
+    }
+    const action = await Hub.findAction(req.actionId, {lookerVersion: req.lookerVersion})
+    return await action.validateAndExecute(request)
+}
+
+process.on("message", (req) => {
+    const response = execute(req)
+    response.then((val) => {process.send!(val)})
```
wilg: nit: could use async?

src/hub/action_request.ts
```diff
@@ -164,6 +164,7 @@ export class ActionRequest {
 
     const streamPromise = new Promise<void>((resolve, reject) => {
       if (url) {
+        process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
```
wilg: should just set this in your local environment for testing if you need it so it doesn't sneak in like this

src/hub/action_request.ts
```diff
@@ -172,7 +173,7 @@ export class ActionRequest {
           })
           .on("finish", resolve)
           .pipe(stream)
-          .on("error", (err) => {
+          .on("error", (err: any) => {
```
wilg: curious this was necessary?

src/server/server.ts
```diff
@@ -110,16 +115,22 @@ export default class Server implements Hub.RouteBuilder {
...
       const request = Hub.ActionRequest.fromRequest(req)
-      const action = await Hub.findAction(req.params.actionId, {lookerVersion: request.lookerVersion})
-      const actionResponse = await action.validateAndExecute(request)
-
-      // Some versions of Looker do not look at the "success" value in the response
-      // if the action returns a 200 status code, even though the Action API specs otherwise.
-      // So we force a non-200 status code as a workaround.
-      if (!actionResponse.success) {
-        res.status(400)
+      const action = await Hub.findAction(req.params.actionId, { lookerVersion: request.lookerVersion })
```
wilg: Can we move all the queue management stuff into its own file/class so that it's not all intermixed with the server code?

src/server/server.ts
```diff
@@ -216,6 +227,33 @@ export default class Server implements Hub.RouteBuilder {
...
+              resolve()
+          }).on("error", (err) => { winston.warn(err.message); reject(err) })
+          const data = {
+              body: req.body,
+              actionId: req.params.actionId,
+              instanceId: req.header("x-looker-instance"),
+              webhookId: req.header("x-looker-webhook-id"),
+              userAgent: req.header("user-agent"),
+          }
+          child.send(data)
```
wilg: the child process imports this as `any`. can we define an explicit interface here that can be shared between the two files.

also, i think it would be prudent to genericize this interface a bit to make it easy to move other jobs into child processes (like the form request, etc.)

src/server/server.ts
wilg: That way we can make the interface to calling the action cross process just look like a single async function call

src/actions/actions_process.ts
phillipperalez: I had to make execute async in order to use await in the message event. If there is a better way to wrap it I would love to hear some suggestions

src/hub/action_request.ts
phillipperalez: ah yes thanks for catching this

src/actions/actions_process.ts
wilg: the `execute` in this file is already async?

you could also not await `action.validateAndExecute` and not make this `execute` async and since it returns a promise it should be awaitable

src/server/server.ts
```diff
@@ -110,16 +115,22 @@ export default class Server implements Hub.RouteBuilder {
...
-      const actionResponse = await action.validateAndExecute(request)
-
-      // Some versions of Looker do not look at the "success" value in the response
-      // if the action returns a 200 status code, even though the Action API specs otherwise.
-      // So we force a non-200 status code as a workaround.
-      if (!actionResponse.success) {
-        res.status(400)
+      const action = await Hub.findAction(req.params.actionId, { lookerVersion: request.lookerVersion })
+      if (action.rateLimited && action.usesStreaming) {
+        await queue.add(async () => {
```
ssirowy: Can be reduced to:
``` jaavscript
await queue.add(async () => this.this.asyncProcess(req, res))
```

src/server/server.ts
```diff
@@ -216,6 +227,33 @@ export default class Server implements Hub.RouteBuilder {
...
+              // Some versions of Looker do not look at the "success" value in the response
+              // if the action returns a 200 status code, even though the Action API specs otherwise.
+              // So we force a non-200 status code as a workaround.
+              winston.info("Success? " + actionResponse.success + " || and what: " + JSON.stringify(actionResponse))
+              if (!actionResponse.success) {
+                  res.status(400)
+              }
+              res.json(actionResponse)
+
+              child.kill()
```
ssirowy: You forgot to kill child in error state too.  Is there a `finally` concept that you can stick this line

src/server/action_process_queue.ts
```diff
@@ -0,0 +1,52 @@
...
+import * as winston from "winston"
+
+export class AsyncProcessJob {
+    queue: Pq
+
+    constructor() {
+        this.queue = new Pq({concurrency: 1})
+    }
+
+    async asyncProcess(req: express.Request, res: express.Response) {
```
wilg: Can we generalize this with something like:

- Making this an abstract class "ExternalProcessJob" which has a `run` function that takes JSON input and returns JSON
- Have a new "ExecuteJob" extend "ExternalProcessJob" and add a new method that translates the action specific stuff into the JSON for XPC
- Leave anything to do with Express in `server.js` since its not really relevant to the execution process or the forking process

src/server/action_process_queue.ts
```diff
@@ -0,0 +1,52 @@
...
+        return this.queue.add(async () => {
+            return new Promise<void>((resolve, reject) => {
+                const child = spawn.fork(`./src/actions/actions_process.ts`)
+                child.on("message", (actionResponse) => {
+                    // Some versions of Looker do not look at the "success" value in the response
+                    // if the action returns a 200 status code, even though the Action API specs otherwise.
+                    // So we force a non-200 status code as a workaround.
+                    winston.info("Success? " + actionResponse.success
+                        + " || and what: " + JSON.stringify(actionResponse))
+                    if (!actionResponse.success) {
```
wilg: that this code is duplicated is a good indication it's not factored as well as it could be. https://github.com/looker/actions/pull/192/files#r211036573 should let you clean that up

src/server/action_process_queue.ts
```diff
@@ -0,0 +1,45 @@
+import * as spawn from "child_process"
+import * as express from "express"
+import * as winston from "winston"
+import {ProcessQueue} from "./process_queue"
+
+export class AsyncProcessJob extends ProcessQueue {
```
wilg: its weird this is a Job but its parent is a Queue

src/server/server.ts
```diff
@@ -110,16 +112,20 @@ export default class Server implements Hub.RouteBuilder {
...
-      // Some versions of Looker do not look at the "success" value in the response
-      // if the action returns a 200 status code, even though the Action API specs otherwise.
-      // So we force a non-200 status code as a workaround.
-      if (!actionResponse.success) {
-        res.status(400)
+      const action = await Hub.findAction(req.params.actionId, { lookerVersion: request.lookerVersion })
+      if (action.expensive && action.usesStreaming) {
+        await expensiveJobQueue.run(req, res)
+      } else {
+        const actionResponse = await action.validateAndExecute(request)
```
wilg: I think we could just move all the fork logic into `validateAndExecute` instead of doing it in the server

wilg: That eliminates the urge to bring Express crap along with us.

src/server/process_queue.ts
```diff
@@ -0,0 +1,11 @@
+import * as Pq from "p-queue"
+
+export abstract class ProcessQueue {
+    queue: Pq
+
+    constructor() {
+        this.queue = new Pq({concurrency: 1})
+    }
+
+    abstract async run(req: Express.Request, res: Express.Response): Promise<void>
```
wilg: I really don't like having this interface involve Express. 

We should have some kind of base job which just knows how to fork a process, send arbitrary data across, and receive data back.

Then on top of that we can build a specialized job which takes an ActionRequest, sends it across to the proper place, performs the execute, and returns the ActionResponse back.

src/hub/action.ts
```diff
@@ -44,6 +44,7 @@ export abstract class Action {
   abstract label: string
   abstract description: string
   usesStreaming = false
+  expensive = false
```
wilg: let's just call this `runInOwnProcess` or something for now

src/server/server.ts
```diff
@@ -110,16 +112,20 @@ export default class Server implements Hub.RouteBuilder {
...
-      const action = await Hub.findAction(req.params.actionId, {lookerVersion: request.lookerVersion})
-      const actionResponse = await action.validateAndExecute(request)
-
-      // Some versions of Looker do not look at the "success" value in the response
-      // if the action returns a 200 status code, even though the Action API specs otherwise.
-      // So we force a non-200 status code as a workaround.
-      if (!actionResponse.success) {
-        res.status(400)
+      const action = await Hub.findAction(req.params.actionId, { lookerVersion: request.lookerVersion })
+      if (action.expensive && action.usesStreaming) {
```
wilg: let's not bother to check both things here – if we say its expensive we'll fork it

src/actions/index.ts
```diff
@@ -8,6 +8,7 @@ import "./google/google_cloud_storage"
 import "./hipchat/hipchat"
 import "./jira/jira"
 import "./marketo/marketo"
+import "./queueaction/queueaction"
```
wilg: Can we name this so its clear its only for testing?

src/actions/queueaction/test_queueaction.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as Hub from "../../hub"
+
+import {ExecuteProcessQueue} from "../../../lib/server/action_process_queue"
+import {QueueAction} from "./queueaction"
+
+const action = new QueueAction()
+const queue = new ExecuteProcessQueue()
+process.execArgv = ["./node_modules/ts-node/dist/bin.js"]
+process.argv = ["node", "./integrations/src/boot.ts"]
+process.env.CHILD_TEST = "true"
```
wilg: how come all these env vars have to be set?

src/server/action_process_queue.ts
```diff
@@ -0,0 +1,29 @@
+import * as spawn from "child_process"
+import * as winston from "winston"
+import {ProcessQueue} from "./process_queue"
+
+export class ExecuteProcessQueue extends ProcessQueue {
+
+    async run(data: string) {
+        return this.queue.add(async () => {
+            return new Promise<string>((resolve, reject) => {
+                const child = spawn.fork(`./src/actions/actions_process.ts`)
```
wilg: can we call this `execute_process` instead for clarity?

src/actions/segment/test_segment_track.ts
```diff
@@ -5,6 +5,7 @@ import * as Hub from "../../hub"
 import { SegmentTrackAction } from "./segment_track"
 
 const action = new SegmentTrackAction()
+action.runInOwnProcess = false
```
wilg: another quick naming thing: let's call this `executeInOwnProcess` for clarity

src/actions/queueaction/test_queueaction.ts
phillipperalez: The process is running with mocha and a different context, it just didn't work without changing the process vars. The process env was just convenient and I saw an example of it being used for testing elsewhere so it made sense

wilg: Do you know what's looking at these env vars? Would just be good to have a comment explaining it if nothing else. Is it something about how `fork` works?

src/actions/queueaction/test_queueaction.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+describe(`${action.constructor.name} unit tests for ProcessQueue`, () => {
+    describe("action", () => {
+        it("returns success on successful response", () => {
+            const request = new Hub.ActionRequest()
+            request.type = Hub.ActionType.Query
+            request.attachment = {dataBuffer: Buffer.from(JSON.stringify({ success: true }))}
+            return expectResponse(request, true)
+        }).timeout(30000)
```
wilg: is actually performing the fork consuming this amount of time?

phillipperalez: Yes it can

src/server/action_process_queue.ts
phillipperalez: I changed it to execute_process_queue . I personally like that name more

src/actions/queueaction/test_queueaction.ts
phillipperalez: Figuring out how to test this was a real pain but at least we have a template for it now

src/actions/queueaction/test_queueaction.ts
phillipperalez: figuring out this stuff was a pain realizing how much has to be made more close to how it runs normally than it would as a test_fixture

src/server/action_process_queue.ts
wilg: actually, its weird that `ExecuteProcessQueue` is in `src/server/action_process_queue.ts` and then there's also `execute_process_queue` now?

Maybe 

- `ExecuteProcessQueue` is in `src/xpc/execute_process_queue.ts` (in `xpc` folder instead of `server`)
- the actual "worker" in in `src/xpc/execute_worker.ts`

src/server/execute_process_queue.ts
```diff
@@ -0,0 +1,29 @@
+import * as spawn from "child_process"
+import * as winston from "winston"
+import {ProcessQueue} from "./process_queue"
+
+export class ExecuteProcessQueue extends ProcessQueue {
+
+    async run(data: string) {
+        return this.queue.add(async () => {
+            return new Promise<string>((resolve, reject) => {
+                const child = spawn.fork(`./src/actions/actions_process.ts`)
```
wilg: I like the rename of this file but I meant renaming `src/actions/actions_process.ts` – perhaps to `src/execute_worker.ts`

wilg: My other comment got outdated by your renaming but before the rename I proposed this https://github.com/looker/actions/pull/192#discussion_r215060258

src/actions/queueaction/queue_action_for_testing.ts
```diff
@@ -0,0 +1,24 @@
+
```
wilg: wonky space (surprised the linter doesn't complain)

wilg:
🎉 :shipit: 🎉

jdavidson:
Incredible!

PR #197, opened 2018-08-23
fix docker file so it will build

rlgomes:
in order to run the yarn command which kicks off the typescript compiler
we must have code already copied into the /code path in the docker
container


rlgomes:
@aabrahamian just look at a fresh checkout and build:
```
> docker build -t actions:latest .
Sending build context to Docker daemon  38.13MB
Step 1/9 : FROM node:10.8.0
 ---> 0e2811757f93
Step 2/9 : RUN npm install -g -s --no-progress yarn@1.9.4
 ---> Using cache
 ---> 3ac8b926c7e7
Step 3/9 : RUN mkdir -p /code
 ---> Using cache
 ---> f39b10918139
Step 4/9 : WORKDIR /code
 ---> Using cache
 ---> d80af18bdbf6
Step 5/9 : COPY package.json yarn.lock tsconfig.json tsconfig.npm.json /code/
 ---> Using cache
 ---> cd3b9ac0f2ce
Step 6/9 : RUN yarn install --production && yarn cache clean
 ---> Running in ef185a914e8f
yarn install v1.9.4
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.2.4: The platform "linux" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
[4/5] Linking dependencies...
warning " > tslint-language-service@0.9.9" has incorrect peer dependency "typescript@>= 2.3.1 < 3".
[5/5] Building fresh packages...
$ ./node_modules/.bin/tsc -d -p tsconfig.npm.json
error TS18003: No inputs were found in config file '/code/tsconfig.npm.json'. Specified 'include' paths were '["src/api_types/**/*.ts","src/hub/**/*.ts","src/server/**/*.ts"]' and 'exclude' paths were '[]'.
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command '/bin/sh -c yarn install --production && yarn cache clean' returned a non-zero code: 1
```

I just don't see how it could work when we run `yarn` before we've put all of the supporting code in place in the container.

You should be able to repro on your own machine simply with:

```
git clone https://github.com/looker/actions
cd actions
docker build -t actions:latest .
```

aabrahamian:
wow, confirmed that this is broken. Digging around it seems that the newer typescript gets upset when there is not at least one .ts file around.
I tried to preserve the caching functionality desired by the change to individual file copies, but couldn't get that to work.

I'll throw some notes on what needs to change and we can get this merged

Dockerfile
```diff
@@ -5,10 +5,10 @@ RUN mkdir -p /code
 WORKDIR /code
 
 COPY package.json yarn.lock tsconfig.json tsconfig.npm.json /code/
```
aabrahamian: let's remove this line in favor of the `COPY . /code` line

Dockerfile
```diff
@@ -5,10 +5,10 @@ RUN mkdir -p /code
...
-RUN yarn install --production && yarn cache clean
-
 COPY . /code
 
+RUN yarn install --production && yarn cache clean
```
wilg: does it work if you drop `--production`?

wilg: nvm

PR #217, opened 2018-10-12
updates Marketo action to use streaming

pascal4mile:
adds a MarketoTransaction class to handle the stateful data needed for the duration of the request.
also adds a Queue class to manage sending 10 requests at a time.


nickfogler:
@wilg I'm not able to make you a reviewer, but I think this is in your court at this point.  Keep in mind that 1) we've not tested outside of Heroku, so our handling of a queue that takes longer than 30 seconds to resolve is theoretical (although we believe it works).  2) We've tried to handle both the specific errors we know about (e.g. record level) and larger errors we have not yet experienced (e.g. API connection failures).  But again, some of this code is as of yet unexercised.  Let us know if it would be useful to have @pascal4mile walk you through any of this.  Cheers!

pascal4mile:
(apologies for the terse commit messages; was having to commit and push a lot in order to test changes on heroku, so i got lazy : )

wilg:
@pascal4mile Looking forward to reviewing this. Can you please fix the tests and update the pull request title and description with an overview of your changes? Thanks!

pascal4mile:
ok @wilg I think this is ready for your eyes. be gentle! no i can take it.

morganatlooker:
Poke @wilg (in case you missed this last week). We have a couple customers waiting on it.

apstrong:
@wilg or @phillipperalez can one of you take a look at this? thanks!

src/actions/marketo/marketo_transaction.ts
```diff
@@ -0,0 +1,236 @@
+/* tslint:disable no-console */
```
phillipperalez: would prefer if logging used Winston instead of console.log

src/actions/marketo/marketo_transaction.ts
```diff
@@ -0,0 +1,236 @@
...
+import * as Hub from "../../hub"
+import { Queue } from "./queue"
+
+const MARKETO: any = require("node-marketo-rest")
+
+const numLeadsAllowedPerCall = 100
+
+// function log(...args: any[]) {
```
phillipperalez: Would just consider taking these comments out or changing to winston for consistency across the repo. @wilg in case he disagrees but I would bet we are on the same page there

src/actions/marketo/queue.ts
```diff
@@ -0,0 +1,135 @@
...
+  counter: number
+  queue: Task[]
+  channels: Task[]
+  completed: Task[]
+  finished: boolean
+  promise: Promise<Task[]>
+  resolve: any
+
+  constructor() {
+    this.channelSize = 10 // TODO make this configurable?
```
phillipperalez: 10 concurrent requests per subscription looks good to me per Marketos docs. One thing that concerns me though is the API rate-limiting. Have you tested a large enough dataset to confirm that we do not send more than 100 API calls per 20 seconds? http://developers.marketo.com/rest-api/marketo-integration-best-practices/

nickfogler: Hey Phillipe, yes our test data set was large enough to test the rate-limiting with no issues, although there's nothing in the code to prevent our hitting the limit if for some reason Marketo's API decides to get super responsive.

pascal4mile:
@wilg @phillipperalez i've updated the action to use winston for logging and removed the commented logging code.

dwmintz:
Awesome! @phillipperalez can you merge?

PR #219, opened 2018-10-30
OAuth Framework and Dropbox

phillipperalez:
Adding the necessary framework to add Oauth enabled actions to the ActionHub


src/crypto/aws_kms.ts
```diff
@@ -0,0 +1,52 @@
+import * as AWS from "aws-sdk"
+import * as winston from "winston"
+import {CryptoBase} from "./crypto_base"
+
+export class AwsKms extends CryptoBase {
```
shaunmistry: Is there a place that you check to ensure that everything required to use AWS KMS is set? Something on the lines of  a check to ensure that the Environment variable pointing to the CMK is set here, and more if required. the check failure can (must) provide useful actionable error message.

wilg:
I don't think I understand two aspects of this design (it seems to have changed since I last understood it)

1. Why is AWS KMS involved here? What function is it performing and why is that required?
2. Why is an application id and secret required to configure this action? I thought a key component of the OAuth design was that the action hub itself would contain the application id and secret so that our customers don't have to individually create Dropbox applications to set up the action.

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,175 @@
...
+            name: "filename",
+            type: "string",
+          }]
+          resolve(form)
+        })
+        .catch((_error: DropboxTypes.Error<DropboxTypes.files.ListFolderError>) => {
+          winston.info("Could not list Dropbox folders")
+          const creds = JSON.stringify({app: request.params.appKey, secret: request.params.secretKey})
+          const kms = new AwsKms()
+          kms.encrypt(creds).then((ciphertextBlob: string) => {
```
wilg: I don't think that individual actions should be manually leveraging particular encryption strategies. Can we move the encryption into an interface and have individual actions leverage encryption through that interface?

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (request.params.state_json) {
+      try {
+        const jsonState = JSON.parse(request.params.state_json)
+        if (jsonState.code && jsonState.redirect) {
+          accessToken = await this.getAccessTokenFromCode(request)
+        }
+      } catch { winston.warn(request.params.state_json) }
+    }
+    const drop = this.dropboxClientFromRequest(request, accessToken)
+    return new Promise<Hub.ActionForm>((resolve, reject) => {
```
wilg: can this use async/await?

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+            })
+            resolve(form)
+          }).catch((err: string) => {
+            winston.error("Encryption not correctly configured")
+            reject(err)
+          })
+        })
+    })
+  }
+  // async oauthUrl(redirectUri: string, stateUrl: string, encryptedState) {
```
wilg: remove comment pls

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+  async oauthUrl(redirectUri: string, encryptedState: string) {
+    const url = new URL("https://www.dropbox.com/oauth2/authorize")
+    const actionCrypto = new Hub.ActionCrypto()
+    const decryptedState = await actionCrypto.decrypt(encryptedState).catch((reason: any) => {
+      throw reason
```
wilg: why the catch and rethrow? won't the `await` just throw?

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+      url: payload.stateurl,
+      body: JSON.stringify({code: urlParams.code, redirect: redirectUri}),
+    }).catch((_err) => { winston.error(_err.toString()) })
+    return `<html><script>window.close()</script>></html>`
+  }
+
+  async oauthCheck(request: Hub.ActionRequest) {
+    let res = false
+    const drop = this.dropboxClientFromRequest(request, "")
+    await drop.filesListFolder({path: ""})
```
wilg: why await AND `then`?

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+      })
+      .catch((error: DropboxTypes.Error<DropboxTypes.files.ListFolderError>) => {
+        winston.error(error.error.toString())
+      })
+    return res
+  }
+
+  protected async getAccessTokenFromCode(request: Hub.ActionRequest) {
+    const url = new URL("https://api.dropboxapi.com/oauth2/token")
+    let stateJson = {code: "something", redirect: "something"}
```
wilg: what is this dummy data?

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+      throw "state_json not present"
+    }
+    url.search = querystring.stringify({
+      grant_type: "authorization_code",
+      code: stateJson.code,
+      client_id: request.params.appKey,
+      client_secret: request.params.secretKey,
+      redirect_uri: stateJson.redirect,
+    })
+    return new Promise<string>((res, _rej) => {
```
wilg: again can we async/await here

src/crypto/aes_transit_crypto.ts
```diff
@@ -0,0 +1,61 @@
+import * as b64 from "base64-url"
+import * as crypto from "crypto"
+import {CryptoBase} from "./crypto_base"
+
+export class AESTransitCrypto extends CryptoBase {
+  ALGORITHM = "aes-256-ctr"
+
+  async encrypt(plaintext: string) {
```
wilg: It's surprising we have such a complicated crypto implementation. Are we doing something special here or can we use something more off-the-shelf?

src/hub/action_form.ts
```diff
@@ -14,7 +20,9 @@ export interface ActionFormField {
   label?: string
   description?: string
   default?: string
-  type?: "string" | "textarea" | "select"
+  type?: "string" | "textarea" | "select" | "set_params_link" | "oauth_link"
```
wilg: Are there situations where we return a form that has fields AND a `"set_params_link"` or  `"oauth_link"`?

If there's no use case for that, it would be a cleaner API to not put these concepts into the `fields` section of the form, but maybe into a new form key that is mutually exclusive with `fields`.

src/hub/action_form.ts
```diff
@@ -14,7 +20,9 @@ export interface ActionFormField {
...
   options?: { name: string, label: string }[]
   required?: boolean
+  oauth_url?: string
+  settable_params?: string[]
```
wilg: If these are only respected when `type` has a particular value, we should encode this into the `ActionFormField` type so the type system will prevent you from specifying them on a `string`.

src/server/server.ts
```diff
@@ -135,6 +135,47 @@ export default class Server implements Hub.RouteBuilder {
...
+      }
+    })
+
+    this.app.get("/actions/:actionId/oauth_check", async (req, res) => {
+      const request = Hub.ActionRequest.fromRequest(req)
+      const action = await Hub.findAction(req.params.actionId, {lookerVersion: request.lookerVersion})
+      if (action instanceof OAuthAction) {
+        const check = action.oauthCheck(request)
+        res.json(check)
+      }
```
wilg: No error condition here?

src/server/server.ts
```diff
@@ -135,6 +135,47 @@ export default class Server implements Hub.RouteBuilder {
       }
     }))
 
+    // Initial OAuth flow request from Resource Owner
+    this.app.get("/actions/:actionId/oauth", async (req, res) => {
+      const request = Hub.ActionRequest.fromRequest(req)
+      const action = await Hub.findAction(req.params.actionId, { lookerVersion: request.lookerVersion })
+      if (action instanceof OAuthAction) {
```
wilg: Should this use `isOauthAction`?

src/server/server.ts
```diff
@@ -135,6 +135,47 @@ export default class Server implements Hub.RouteBuilder {
...
+        const check = action.oauthCheck(request)
+        res.json(check)
+      }
+    })
+
+    // Response from Authorization Server with response from OAuth flow
+    this.app.get("/actions/:actionId/oauth_redirect", async (req, res) => {
+      const request = Hub.ActionRequest.fromRequest(req)
+      const action = await Hub.findAction(req.params.actionId, { lookerVersion: request.lookerVersion })
+      if (action instanceof OAuthAction) {
```
wilg: Should this use `isOauthAction`?

src/server/server.ts
```diff
@@ -135,6 +135,47 @@ export default class Server implements Hub.RouteBuilder {
...
+        res.redirect(url)
+      } else {
+        throw "Action does not support OAuth."
+      }
+    })
+
+    this.app.get("/actions/:actionId/oauth_check", async (req, res) => {
+      const request = Hub.ActionRequest.fromRequest(req)
+      const action = await Hub.findAction(req.params.actionId, {lookerVersion: request.lookerVersion})
+      if (action instanceof OAuthAction) {
```
wilg: Should this use `isOauthAction`?

src/server/server.ts
```diff
@@ -151,6 +192,10 @@ export default class Server implements Hub.RouteBuilder {
     return this.absUrl(`/actions/${encodeURIComponent(action.name)}/form`)
   }
 
+  oauthRedirectUrl(action: Hub.Action) {
```
wilg: should this be private?

src/server/server.ts
```diff
@@ -190,7 +235,7 @@ export default class Server implements Hub.RouteBuilder {
 
       const headerValue = req.header("authorization")
       const tokenMatch = headerValue ? headerValue.match(TOKEN_REGEX) : undefined
-      if (!tokenMatch || !apiKey.validate(tokenMatch[1])) {
+      if (!tokenMatch || !apiKey.validate(tokenMatch[1]) && urlPath !== "/actions/:actionId/oauth_redirect") {
```
wilg: Maybe instead we can have a `skipAuthentication` param to `route`?

tslint.json
```diff
@@ -16,6 +16,7 @@
       "no-inferred-empty-object-type": true,
       "strict-type-predicates": true,
       "strict-boolean-expressions": [true, "allow-undefined-union", "allow-null-union", "allow-mix", "allow-string"],
+      "triple-equals": [true, "allow-null-check"],
```
wilg: What do we need this for?

bin/generate_key.ts
```diff
@@ -1,9 +1,8 @@
+require("dotenv").config()
```
wilg: What's the idea here? To get this before `crypto` is required? I think it's a bit scary because `import` is supposed to come before any application code and the specific way we are compiling typescript might be the only reason this is allowed.

I think you can actually do `import 'dotenv/config'` according to the docs (https://www.npmjs.com/package/dotenv) to get around this.

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+    label = "Dropbox"
+    iconName = "dropbox/dropbox.png"
+    description = "Send query results directly to a file in your Dropbox."
+    supportedActionTypes = [Hub.ActionType.Query, Hub.ActionType.Dashboard]
+    usesStreaming = false
+    minimumSupportedLookerVersion = "6.4.0"
+    requiredFields = []
+    params = [
+      {
+        description: "Dropbox Application Key",
```
wilg: I'm still going to be a pest about trying to make it so that users don't need to provide this key/secret by having us own the Dropbox app.

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,211 @@
...
+      if (jsonState.code && jsonState.redirect) {
+        accessToken = await this.getAccessTokenFromCode(request)
+      }
+    }
+    const drop = this.dropboxClientFromRequest(request, accessToken)
+
+    const resp = new Hub.ActionResponse()
+    if (request.attachment && request.attachment.dataBuffer) {
+      const fileBuf = request.attachment.dataBuffer
+      await drop.filesUpload({path: `/${directory}/${filename}.${ext}`, contents: fileBuf}).then((_dropResp) => {
```
wilg: Why await AND then?

src/actions/dropbox/dropbox.ts
phillipperalez: I mean it can - I personally liked this approach for this section though. It made more sense to me as I was writing it

src/hub/action_form.ts
phillipperalez: oh good catch, "set_params_link" is cruft that needs to be removed

bin/generate_key.ts
phillipperalez: oh that's from the original PR as well. deleting that

src/actions/dropbox/dropbox.ts
wilg: Interesting... typically I think async/await is easier to understand versus chained callbacks.

src/actions/dropbox/dropbox.ts
phillipperalez: fixed

src/actions/dropbox/dropbox.ts
JimRottinger: I'm guessing the linter requires the catch? That would be my guess but Wil is right that this is just catching and throwing immediately

phillipperalez: linter does require the catch. Adding more information here to make a better message

src/hub/oauth_action.ts
```diff
@@ -0,0 +1,13 @@
...
+import {ActionRequest} from "./action_request"
+
+export abstract class OAuthAction extends Action {
+  usesOauth = true
+  abstract async oauthCheck(request: ActionRequest): Promise<boolean>
+  abstract async oauthUrl(redirectUri: string, encryptedState: string): Promise<string>
+  abstract async oauthFetchInfo(urlParams: { [key: string]: string }, redirectUri: string): Promise<string>
+}
+
+export function isOauthAction(action: Action): boolean {
```
JimRottinger: I would think this function belongs on the action itself instead of as a separate export or does not violate the statelessness?

src/server/server.ts
```diff
@@ -151,6 +193,10 @@ export default class Server implements Hub.RouteBuilder {
     return this.absUrl(`/actions/${encodeURIComponent(action.name)}/form`)
   }
 
+  private oauthRedirectUrl(action: Hub.Action) {
+    const url = this.absUrl(`/actions/${encodeURIComponent(action.name)}/oauth_redirect`)
+    return url
```
JimRottinger: Just return the call to `this.absUrl` unless you're expecting to do something else with it.

JimRottinger: Can't hurt to add return types (`string`) to both of the functions too

wilg: the return type should be inferred

src/actions/dropbox/dropbox.ts
jdavidson: @phillipperalez is there a need for the user to create their own dropbox app? it seems simpler for a single action hub dropbox app to be created.

phillipperalez: I need to talk with dropbox. It's a relatively easy change for me to make if they can just make us a global app but that isn't always a guarantee

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+      } catch { winston.warn("Could not parse state_json") }
+    }
+    const drop = this.dropboxClientFromRequest(request, accessToken)
+    try {
+      const response = await drop.filesListFolder({path: ""})
+      let folderList = response.entries.filter((entries) => (entries[".tag"] === "folder"))
+        .map((entries) => ({name: entries.name, label: entries.name}))
+      if (folderList.length === 0) {
+        folderList = [{name: "Looker", label: "Looker"}]
```
wilg: What is this about?

src/boot.ts
```diff
@@ -1,3 +1,5 @@
+require("dotenv").config()
```
wilg: Can we use the import style here? `import 'dotenv/config'`

wilg: Just because having code that runs before `import` statements is not really part of the JS standard I think

src/crypto/crypto_base.ts
```diff
@@ -0,0 +1,4 @@
+export abstract class CryptoBase {
```
wilg: Nit: This should probably be an interface instead of an abstract class since it has no implementation. Additionally, we should probably make sure that `ActionCrypto` is typed with the interface `CryptoBase` versus having the concrete type accessible to actions. Also maybe we should rename this something more like `CryptoProvider`.

src/actions/dropbox/dropbox.ts
phillipperalez: if there are no valid folders to list than do not use "undefined" as the folder name. Actually give a fresh dropbox an option

wilg: This folder list is a bit wonky then, because it will only show immediate subdirectories of your Dropbox folder, right? I know we don't have a good form option for dealing with hierarchical data. I do think this is a bit confusing though, but I don't know the right answer.

src/hub/action_response.ts
```diff
@@ -24,18 +27,17 @@ export class ActionResponse {
 
   asJson(): any {
     const errs: any = {}
-    if (this.validationErrors) {
+    if (this.validationErrors.length > 0) {
```
wilg: I wonder why originally it was possibly `undefined`?

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,187 @@
...
+      winston.error("Encryption not correctly configured" + err)
+      throw err
+    })
+
+    const payload = JSON.parse(plaintext)
+    await https.post({
+      url: payload.stateurl,
+      body: JSON.stringify({code: urlParams.code, redirect: redirectUri}),
+    }).catch((_err) => { winston.error(_err.toString()) })
+    return `<html><script>window.close()</script>></html>`
```
wilg: What's the use case for returning something other than this? Kind of annoying/scary for every action to be returning arbitrary HTML/JS. Can we standardize this?

src/hub/action_form.ts
```diff
@@ -14,7 +20,8 @@ export interface ActionFormField {
   label?: string
   description?: string
   default?: string
-  type?: "string" | "textarea" | "select"
+  type?: "string" | "textarea" | "select" | "oauth_link"
```
wilg: If these are only respected when type has a particular value, we should encode this into the ActionFormField type so the type system will prevent you from specifying them on a string.

src/actions/dropbox/dropbox.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  async form(request: Hub.ActionRequest) {
+    const form = new Hub.ActionForm()
+    form.fields = []
+
+    let accessToken = ""
+    if (request.params.state_json) {
+      try {
+        const jsonState = JSON.parse(request.params.state_json)
```
wilg: This parsing appears to happen twice, once here and the other in `getAccessTokenFromCode`. Can that be deduplicated?

phillipperalez: yeah - I forgot to remove it when I took the key and secret out of the params field

src/actions/dropbox/dropbox.ts
phillipperalez: Need to auto close the new tab that was opened for oauth with Dropbox's authorization server. Seemed like the cleanest way to do it

wilg: What I meant to say is can we push the scary `<html><script>window.close()</script>></html>` up into the server or base class or somewhere? Would you expect other actions to need this?

src/hub/action_form.ts
wilg: This will make the TS compiler errors enforce the required properties for each `type` of field.

```ts
export interface ActionFormFieldBase {
  name: string
  label?: string
  description?: string
  default?: string
  required?: boolean
}

export interface ActionFormFieldString extends ActionFormFieldBase {
    type?: "string" | "textarea"
}

export interface ActionFormFieldSelect extends ActionFormFieldBase {
    type: "select" 
    options: { name: string, label: string }[]
}

export interface ActionFormFieldOAuth extends ActionFormFieldBase {
  type: "oauth_link" 
  oauth_url: string
}

export type ActionFormField = ActionFormFieldString | ActionFormFieldSelect | ActionFormFieldOAuth
```

PR #222, opened 2018-11-26
lock event-stream to non-compromised version 3.3.4 - Prod unaffected // Test/Dev Security Fix

phillipperalez:
from: https://github.com/dominictarr/event-stream/issues/115

Dependency libraries should have already updated this but let's make sure that the version is at a non-compromised version explicitly. This did not affect production as this is a ` nodemon ` vulnerability. We do not run nodemon for prod, only for test/dev commands.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1314,9 +1314,9 @@ duplexer3@^0.1.4:
   version "0.1.4"
   resolved "https://registry.yarnpkg.com/duplexer3/-/duplexer3-0.1.4.tgz#ee01dd1cac0ed3cbc7fdbea37dc0a8f1ce002ce2"
 
-duplexer@~0.1.1:
+duplexer@^0.1.1, duplexer@~0.1.1:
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1401,6 +1401,18 @@ etag@~1.8.1:
...
+  dependencies:
+    duplexer "^0.1.1"
+    from "^0.1.7"
+    map-stream "0.0.7"
+    pause-stream "^0.0.11"
+    split "^1.0.1"
+    stream-combiner "^0.2.2"
+    through "^2.3.8"
+
 event-stream@~3.3.0:
```
wilg: The old version of event-stream is still here, just adding a top-level dependency in NPM doesn't mean that it will be used by other dependencies, since multiple versions can be in play at the same time.

I think you'll need to use the `resolutions` feature: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

PR #223, opened 2018-11-28
Add url field to DataRobot action

pavloko:
We have to update the DatRobot Action to support enterprise customers who might have their instance hosted under arbitrary URL.

I've also added a README describing the setup process.


pavloko:
@wilg for some reason my test behave in a weird way; despite not mocking http requests, they still pass. I've also experimented with other action tests and had similar results.

Am I missing something?

pavloko:
fixed conflicts all conflicts, any chance this gets merged?

PR #224, opened 2018-11-29
Adding json-detail-lite-stream in the case of Segment and Looker v6.2 or greater

phillipperalez:
This format is not 100% the same as json_detail  however it provides what Segment needs in order to stream in an unlimited fashion.

### DIFF
--- a/src/actions/segment/segment.ts
+++ b/src/actions/segment/segment.ts
@@ -2,8 +2,11 @@ import * as util from "util"
 import * as uuid from "uuid"
 import * as winston from "winston"
 
+import * as semver from "semver"
 import * as Hub from "../../hub"
-import { SegmentActionError } from "./segment_error"
+import {ActionDownloadSettings} from "../../hub"
+import {RouteBuilder} from "../../hub"
+import {SegmentActionError} from "./segment_error"
 
 const segment: any = require("analytics-node")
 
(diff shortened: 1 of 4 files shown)


src/actions/segment/segment.ts
```diff
@@ -55,6 +58,34 @@ export class SegmentAction extends Hub.Action {
   requiredFields = [{ any_tag: this.allowedTags }]
   executeInOwnProcess = true
 
+  asJson(router: RouteBuilder, request?: any): any {
```
wilg: This is a repetition of the entire `asJson` implementation from the superclass right? That seems extremely brittle. 

What if instead we did something like `supportedFormats` could optionally be a function like `this.supportedFormats = (request) => { ... }` that can modulate the supported formats based on the current request?

src/actions/marketo/marketo.ts
```diff
@@ -31,10 +32,16 @@ export class MarketoAction extends Hub.Action {
 
   ]
   supportedActionTypes = [Hub.ActionType.Query]
-  supportedFormats = [Hub.ActionFormat.JsonDetail]
   supportedFormattings = [Hub.ActionFormatting.Unformatted]
   supportedVisualizationFormattings = [Hub.ActionVisualizationFormatting.Noapply]
   usesStreaming = true
+  supportedFormats = (request: Hub.ActionRequest) => {
```
wilg: I would write a quick test for these two actions that validates the behavior on different looker versions

phillipperalez: good idea

PR #231, opened 2019-03-07
adds Amazon Sagemaker actions

pascal4mile:


src/actions/sagemaker/test_sagemaker.ts
```diff
@@ -0,0 +1,44 @@
+// import * as chai from "chai"
```
pascal4mile: nothing to see here yet, need to figure out how to run tests, need to figure out what needs to be tested. pointers appreciated

src/actions/sagemaker/sagemaker_infer.ts
```diff
@@ -0,0 +1,346 @@
...
+      winston.debug("transformParams", transformParams)
+
+      const transformResponse = await client.createTransformJob(transformParams).promise()
+      winston.debug("transformResponse", transformResponse)
+
+      // return success response
+      return new Hub.ActionResponse({ success: true })
+
+    } catch (err) {
+      winston.error(err)
```
phillipperalez: For anything that is going to be logged at error or higher - the ` request.webhookId ` is a nice thing to add for debugging. That way we can track when and how a request was made in Looker and if there were any errors constructing the it

src/actions/sagemaker/sagemaker_infer.ts
```diff
@@ -0,0 +1,346 @@
...
+        name: "bucket",
+        required: true,
+        options: buckets.map((bucket) => {
+          return {
+            name: bucket.Name!,
+            label: bucket.Name!,
+          }
+        }),
+        type: "select",
+        // default: buckets[0].Name, // DNR
```
phillipperalez: extraneous comment?

src/actions/sagemaker/test_sagemaker.ts
pascal4mile: @phillipperalez thanks for the review. I've address the issues you raised below. I'm a bit stumped on how to write tests against LookerActions. 
- at a high level, I sort of get what needs to be tested: looks like most actions are testing the execute and form methods, asserting the errors returned with bad inputs, and asserting the response with good inputs.
- at a lower level, i'm kinda stuck on implementation details like stubbing out the AWS client, and getting some errors like `Error: Invalid Chai property: eventually`.

Could I get some help on that?

phillipperalez: Hi Pascal. Sorry for the very late response. The best way for these unit tests that I have found is through using a separate function to set up the execute and stubbing all of the outside calls in there. 
for example in https://github.com/looker/actions/blob/b469ccca2d604bec2fb42a7a59691251649c9739/src/actions/dropbox/test_dropbox.ts#L51-L74
I made a function that stubbed function that creates the client as well as the call to that client that would make an API call. At the end of the test I always remember to restore what I have stubbed so the next test is not affected.
https://github.com/looker/actions/blob/b469ccca2d604bec2fb42a7a59691251649c9739/src/actions/dropbox/test_dropbox.ts#L82-L85
The short and sweet - stub a function to return a default ` action ` (created at the top of the test file) and create expectations for the functions you expect to be called on ` action `

phillipperalez: https://github.com/looker/actions/blob/b469ccca2d604bec2fb42a7a59691251649c9739/src/actions/dropbox/test_dropbox.ts#L136-L143
Here is an example of where I return something from a stubbed call that would have made an API request and instead return something which the Dropbox form method expects on success

src/actions/sagemaker/training_job_poller.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.transporter = getMailTransporter(transaction.request)
+    this.pollTrainingJob(transaction).catch(logRejection)
+  }
+
+  async pollTrainingJob(transaction: Transaction) {
+      // start poller for training job completion
+      winston.debug("starting poller")
+
+      this.intervalTimer = setInterval(() => {
+        this.checkTrainingJob(transaction).catch(logRejection)
```
pascal4mile: Got a lot of lint errors about not handling promises correctly. As a fix I added these .catch handlers to log the rejection. Not sure if this is an adequate solution.

pascal4mile:
@phillipperalez I've added the tests, so I think this is ready for another review.

PR #235, opened 2019-03-26
Let's Update the Action Hub API Docs

phillipperalez:
Work in progress update of all docs on ActionHub

### DIFF
--- a/docs/action_api.md
+++ b/docs/action_api.md
@@ -75,7 +75,7 @@ interface ActionDefinition {
   description?: string
   /** Array of params for the action. */
   params?: Param[]
-  /** A list of data formats the action supports. Valid values are: "txt", "csv", "inline_json", "json", "json_detail", "xlsx", "html", "wysiwyg_pdf", "assembled_pdf", "wysiwyg_png". */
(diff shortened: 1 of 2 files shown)


docs/action_api.md
```diff
@@ -203,6 +219,19 @@ interface FormSelectOption {
 }
 ```
 
+### Action User State
+
+By default the Action Hub keeps each an every action independent from previous actions. For certain actions though (OAuth enabled actions being an example) - state may be required per user. If there is state for a user and action - then Looker will send the state as part of a `form` or `execute` request. If there is no state then the field is unused. If you wish to set state it can be accomplished using the `state` field on an `execute` or `form` response. Refresh time may also be specified in seconds (a refresh on state will make a state request), although the Looker server currently will only look for new refresh states once every 10 minutes. This can be used to keep authentication credentials up to date or to simply keep some state up to date on the Looker server for the user and action.
```
wilg: typo "each an every"

wilg: I would also maybe change "By default the Action Hub" to "The action hub API is designed to"

PR #237, opened 2019-04-02
Address Branding Issues with Dropbox

phillipperalez:


src/actions/dropbox/dropbox.ts
```diff
@@ -128,6 +128,7 @@ export class DropboxAction extends Hub.OAuthAction {
     })
 
     const payload = JSON.parse(plaintext)
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
```
wilg: wat

wilg: (since this project uses `dotenv`, you can just put `NODE_TLS_REJECT_UNAUTHORIZED=0` in a `.env` file at the root)

phillipperalez: oh whoops - just test code I use that accidentally made it up. thanks for the catch

PR #242, opened 2019-04-08
Remove Hipchat as they no longer exist

phillipperalez:


jdavidson:
Thanks this sounds good. Couple quick questions:

Is Hipchat now fully deprecated? https://www.atlassian.com/partnerships/slack seems to imply there is a migration path.
What would happen to any existing schedules that have previously connected to hipchat?

Thanks!

PR #243, opened 2019-04-18
Fabio looker/marketo endpoint enhancements

fabio-looker:
PR adds additional modes to the Marketo action, specifically:
- Being able to update leads WITHOUT providing a campaign ID
- Being able to update leads and add them to a list
- Being able to update leads and remove them from a list

image

### DIFF
--- a/src/actions/marketo/marketo_transaction.ts
+++ b/src/actions/marketo/marketo_transaction.ts
@@ -26,40 +26,44 @@ export class MarketoTransaction {
     this.campaignIds = []
     this.addListIds = []
     this.removeListIds = []
-    let subactionIds = (request.formParams.subactionIds || '').split(/\s*,\s*/).filter(Boolean)
-    switch(request.formParams.subaction){
+    const subactionIds =
+      (request.formParams.subactionIds !== undefined ? request.formParams.subactionIds : "")
+      .split(/\s*,\s*/)
+      .filter(Boolean)
+    switch (request.formParams.subaction) {
       case undefined:
-        //The older version of the action assumed an "addToCampaign" subaction
-        this.campaignIds = [request.formParams.campaignId || ''].filter(Boolean) //Note singular parameter name
-        break;
-      case 'none':
-        if(subactionIds.length){
+        // The older version of the action assumed an "addToCampaign" subaction
+        this.campaignIds = [request.formParams.campaignId !== undefined ? request.formParams.campaignId : ""]
+          .filter(Boolean) // Note singular parameter name
+        break
+      case "none":
+        if (subactionIds.length === 0) {
           throw "Additional action of 'none' was selected, but additional action ID was provided"
         }
-      break;
-      case 'addCampaign':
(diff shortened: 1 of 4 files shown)


fabio-looker:
Fixing TS lint stuff...

dwmintz:
cc: @phillipperalez

src/actions/marketo/marketo_transaction.ts
```diff
@@ -10,21 +10,58 @@ interface Result {
...
+    const subactionIds =
+      (request.formParams.subactionIds !== undefined ? request.formParams.subactionIds : "")
+      .split(/\s*,\s*/)
+      .filter(Boolean)
+    switch (request.formParams.subaction) {
+      case undefined:
+        // The older version of the action assumed an "addToCampaign" subaction
+        this.campaignIds = [request.formParams.campaignId !== undefined ? request.formParams.campaignId : ""]
+          .filter(Boolean) // Note singular parameter name
+        break
```
dwmintz: @phillipperalez this logic is critical to make sure that existing schedules that were created with the existing version of the Marketo action do not break.  It defaults to the update record and add to campaign action, which is the only one that the current version supports.

apstrong:
@jdavidson or @wilg could one of you review this? Phillip's focusing on alerts work this week

jdavidson:
I'll try to take a look in a bit. Why not make each of the four different behaviors 4 different actions?

fabio-looker:
@jdavidson No particular reason. It just seemed like a pretty big code change to make separate actions for each.

dwmintz:
I strongly prefer having them all in one from a UI perspective. Just feels like it'd be overwhelming in the picker to start having many of each logo

Best,
-d

fabio-looker:
Please hold on review until tomorrow. I need to fix bug that I introduced during my "fixing" of Typescript linter warnings. And also the test that I thought should catch this case should be expanded...

fabio-looker:
Btw, this should now be good for review

dwmintz:
Just pinging @wilg and @jdavidson on this to see if one of you can take a look.

maxreid1:
I have a prospect in trial very interested in this - do we have an ETA on it?

dwmintz:
Approving the readme changes.

dwmintz:
@fabio-looker we did some testing of this on dev, and while generally it looks great, there's one bug that I think we need to fix before releasing. When we upgraded from v1 to v2, we used a test schedule to make sure that it didn't break, and unfortunately it did :-(

The behavior I saw was that upon upgrade, the existing schedule correctly kept the custom lookup field that I'd entered for the saved schedule, but it lost the campaign_id (it went to blank in v2) and the dropdown defaulted to `none`, rather than `addCampaign` (which it will always be if it's coming from v1).

Is that something you can take a look at?

fabio-looker:
Certainly. I'm at the hackathon today. If I take a look by Friday is that alright?

dwmintz:
Yup! Thanks.

Best,
-d

PR #249, opened 2019-05-23
Alphabetize destination dropdown for Slack

ryandunlavy:
I just added this since I noticed the lookerbot action had alphabetization for the dropdown for Slack but the action didn't https://github.com/looker/lookerbot/blob/master/src/services/slack_service.ts#L132

This will sort the list of channels and list of users separately before merging the two lists.

### DIFF
--- a/src/actions/slack/slack.ts
+++ b/src/actions/slack/slack.ts
@@ -3,7 +3,6 @@ import * as Hub from "../../hub"
 import { WebClient } from "@slack/client"
 import * as _ from 'underscore'
 
-
 interface Channel {
   id: string,
   label: string,
--- a/src/actions/slack/slack.ts
+++ b/src/actions/slack/slack.ts
@@ -1,7 +1,6 @@
 import * as Hub from "../../hub"
 
 import { WebClient } from "@slack/client"
-import * as _ from 'underscore'
 
 interface Channel {
   id: string,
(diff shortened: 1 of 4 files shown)


src/actions/slack/slack.ts
```diff
@@ -124,7 +125,7 @@ https://github.com/looker/actions/blob/master/src/actions/slack/README.md`,
     const paginatedChannels = await pageLoaded([], await slack.channels.list(options))
     const channels = paginatedChannels.filter((c: any) => c.is_member && !c.is_archived)
     const reformatted: Channel[] = channels.map((channel: any) => ({id: channel.id, label: `#${channel.name}`}))
-    return reformatted
+    return _.sortBy(reformatted, "label")
```
jdavidson: we can probably pull this into the form in the definition of the fields and do 1 sort.

also, can we just use js vs. including underscore?

`channels.sort((a, b) => (a.label.localeCompare(b.label)))`

ryandunlavy: Ah nice, this is much better. This would make DMs appear before public channels but maybe something like this could work

```
channels.sort((a, b) =>  (a.label.startsWith("#") && b.label.startsWith("@") ? -1 : a.label.localeCompare(b.label)))
```

ryandunlavy: It ends up checking for order with > sorts # and @ the desired way so I just used that

PR #253, opened 2019-06-20
Add looker key that lookers expect

jdavidson:


wilg:
OK this is weird since I did this a year ago https://github.com/looker/actions/pull/173 – clearly we are missing a test around this.

wilg:
Almost exactly a year ago, actually.

PR #255, opened 2019-06-25
Unified slack action w/ optional oauth

jdavidson:


phillipperalez:
New slack app is what will be used

PR #257, opened 2019-07-02
Slack can only send to public channels

Kruncha:
According to this FR: https://github.com/looker/analytics/issues/7362


animcdowell:
This is updated now! https://github.com/looker/actions/blob/master/src/actions/slack/legacy_slack/README.md

animcdowell:
This is updated now! https://github.com/looker/actions/blob/master/src/actions/slack/legacy_slack/README.md

PR #261, opened 2019-08-06
add auger action

derekflint:
Auger integration request.


src/actions/auger/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 Looker Data Sciences, Inc.
+
```
animcdowell: @derekflint can you add a second line here after Looker's for Auger.AI? Example: `Copyright (c) 2019 Auger.AI`

phillipperalez:
I just added a second extended action queue to help make sure this action does not block other heavy actions. This can be enabled by adding ` extendedAction = true `

src/actions/auger/auger_train.ts
```diff
@@ -0,0 +1,488 @@
...
+export class AugerTrainAction extends Hub.Action {
+  name = "auger"
+  label = "Auger - AutoML Training"
+  iconName = "auger/auger.png"
+  description = "Send data to Auger to start training."
+  supportedActionTypes = [Hub.ActionType.Query]
+  supportedFormattings = [Hub.ActionFormatting.Unformatted]
+  supportedVisualizationFormattings = [Hub.ActionVisualizationFormatting.Noapply]
+  requiredFields = []
+  usesStreaming = true
```
phillipperalez: Please add `executeInOwnProcess=true` and `extendedAction = true` below here to make sure that it executes in a different queue.

src/actions/auger/test_auger_train.ts
```diff
@@ -0,0 +1,228 @@
+import * as chai from "chai"
+import * as httpRequest from "request-promise-native"
+import * as sinon from "sinon"
+// import * as winston from "winston"
+import * as Hub from "../../hub"
+import { Transaction } from "./poller"
+
+import { AugerTrainAction } from "./auger_train"
+
+const action = new AugerTrainAction()
```
phillipperalez: use `action.executeInOwnProcess = false` to make sure we are just testing your action and not the process queueing mechanisms

src/actions/auger/test_auger_train.ts
```diff
@@ -0,0 +1,228 @@
+import * as chai from "chai"
+import * as httpRequest from "request-promise-native"
+import * as sinon from "sinon"
+// import * as winston from "winston"
```
phillipperalez: nit: remove this comment

src/actions/auger/auger_train.ts
```diff
@@ -0,0 +1,490 @@
...
+        json: true,
+        resolveWithFullResponse: true,
+      }
+      return httpRequest.get(options).promise()
+    } catch (e) {
+      throw new Error(`project file fetched: ${e}`)
+    }
+  }
+
+  // private async getExperiment(transaction: Transaction): Promise<any> {
```
phillipperalez: Is this commented function needed anymore?

derekflint: I'll remove it

phillipperalez:
Alright! I made a new way for Auger.AI to execute without interfering with the other queues. This was implemented in https://github.com/looker/actions/pull/289

The idea was that Auger is doing a long polling/async chain of work that would not work in the current `executeInOwnProcess` queue. This is because the old queue would close the process as soon as the action responded back to Looker. Obviously this doesn't work with the Auger.AI action so I made a new process queue that can both return back to Looker and keep its process alive until it is done. I am in the process of writing more docs for that but they won't be done fast enough so I am going to explain how to use it here

to make an "extendedAction", add both
```
  executeInOwnProcess = true
  extendedAction = true
```
to your action
the code that returns a response back to Looker is still fine and works as intended. There is one final thing that needs to be done though to make sure that the process closes when the work is actually done. 
```
process.send("PROCESS FINISHED")
```
somewhere at the very end the action needs to send the above message back to the parent process to let it know that its work is done. For testing I used a couple actions and wrapped the call in an async function 
```
  async callback() {
    if (process.send) {
      process.send("PROCESS FINISHED")
    }
  }
```
and then just passed that callback into whatever needed to call it. Please let me know if any of that is confusing, this is a new concept for the Action Hub

derekflint:
@phillipperalez from what I understand I should add `process.send("PROCESS FINISHED")` inside my action when its finished with all async calls. 

I tried this but get error `Cannot invoke an object which is possibly 'undefined'.`

I've tried wrapping call in !== undefined but same issue. Do I need to import anything?

phillipperalez:
Did you pull latest master? On your fork it looks like it is a few commits behind. Also this callback (the name of the function is something you can of course change)
```
  async callback() {
    if (process.send) {
      process.send("PROCESS FINISHED")
    }
  }
```
This should work and I have tested it again

derekflint:
I've pulled and fixed the conditional check.

PR #265, opened 2019-08-20
Oauth Jira Integration

jdavidson:


src/actions/jira/jira.ts
```diff
@@ -1,133 +1,207 @@
...
-        required: true,
-      }, {
-        default: issueTypes[0].id,
-        label: "Issue Type",
-        name: "issueType",
-        type: "select",
-        options: issueTypes
-          .filter((i: any) => i.description)
-          .map((p: any) => {
+    const actionCrypto = new Hub.ActionCrypto()
```
phillipperalez: shouldn't the oauth link be if the login doesn't work and not all the time?

jdavidson: now addressed

PR #271, opened 2019-09-25
Braze Action - Add optional Attributes to Action Hub

zzhaobraze:
Added a feature to allow for sending optional custom and standard Braze attributes via tagging with `braze[]`. 
Also increased api limit from 10,000 to 100,000.


src/actions/braze/braze.ts
```diff
@@ -96,14 +101,24 @@ export class BrazeAction extends Hub.Action {
...
     let rows: BrazeApiRow[] = []
+    const bzRegExp = new RegExp(BrazeConfig.BRAZE_ATTRIBUTE_REGEX, "gi")
     try {
       await request.streamJsonDetail({
         onFields: (fields) => {
           fieldlist = Hub.allFields(fields)
           for (const field of fieldlist) {
-            if (field.tags  && field.tags.find((tag: string) => tag === BrazeConfig.BRAZE_ID_TAG )) {
-              bzIdField = field.name
+            if (field.tags) {
```
CCedricYoung: nit - should we use a guard clause instead?
https://blog.codinghorror.com/flattening-arrow-code/
```suggestion
            if (!field.tags) {
              continue
            }
```

src/actions/braze/braze.ts
```diff
@@ -96,14 +101,24 @@ export class BrazeAction extends Hub.Action {
...
+              if (field.tags.find((tag: string) => tag === BrazeConfig.BRAZE_ID_TAG )) {
+                bzIdField = field.name
+              }
+              for (const tag of field.tags) {
+                const bzTagMatchs = tag.match(bzRegExp)
```
CCedricYoung: nit - spelling - should this be `bzTagMatches`?

zzhaobraze: changed

src/actions/braze/braze.ts
zzhaobraze: i don't have a specific preference so this has been changed.

phillipperalez:
https://github.com/looker/actions/pull/271/files#diff-c5149b4c257c672d74b56fbc8a3d3813R51
I would like `executeInOwnProcess = true` here. Reason being is that we are now adding some more processing on an unlimited amount of data, getting some potentially very large streams of data off of the main event loop would probably be a good idea

example usage: https://github.com/looker/actions/blob/master/src/actions/segment/segment.ts#L56

phillipperalez:
looked at the test failure. all that needs to happen is in the Braze test file, use ` action.executeInOwnProcess = false ` like in `test_segment.ts`. We test the queueing in another file and in the Braze action we just want to test the Braze action itself without the queueing

zzhaobraze:
> looked at the test failure. all that needs to happen is in the Braze test file, use `action.executeInOwnProcess = false` like in `test_segment.ts`. We test the queueing in another file and in the Braze action we just want to test the Braze action itself without the queueing

Thanks. Made the change.

PR #272, opened 2019-09-26
Slack OAuth

annguy3n:
image

This PR allows user to add slack integration with OAuth action in AH side, there is corresponding HT PR https://github.com/looker/helltool/pull/47078


phillipperalez:
I have been reviewing the full stack required to make this PR work as well as this PR and I think it is good to go

PR #275, opened 2019-10-01
Mparticle custom action

jj57347:
Requires API key and secret to Mparticle's batch API
Takes either user data or event data and batches them to their API


jj57347:
To whoever is reviewing this PR - I noticed that for large (5000+ rows) action execution, the action performs fine and sends all the data appropriately, but because the streaming takes longer than Looker's timeout window, the user still receives a 404 email indicating that the action did not complete successfully. Is this a bug on Looker's side? I am returning `new Hub.ActionResponse({ success: true })` in the same way that I see other actions doing.

phillipperalez:
```
  MparticleAction unit tests
    action
      ✓ errors if there is no apiKey
      ✓ errors if there is no apiSecret
      ✓ errors if there is no data_type
      1) errors if there is not at least one userIdentity
    asJSON
      ✓ supported format is json_detail_lite_stream on lookerVersion 6.2 and above
```
Just this one error left it looks like in your tests

jj57347:
@phillipperalez Hmm that is odd, it was working before and I haven't changed that flow - I see it working as expected in the action itself, do you see anything about the test or error throwing that might be weird? I tried to keep to the same style as I saw in other actions

phillipperalez:
Running the test locally it doesn't seem like the `Each row must specify at least 1 identity tag.` is getting thrown with the sample data. Looking into it too

src/actions/mparticle/mparticle_transaction.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  async sendChunk(rows: Hub.JsonDetail.Row[], mapping: any) {
+    const chunk = rows.slice(0)
+    const body: MparticleBulkEvent[] = []
+    chunk.forEach((row: Hub.JsonDetail.Row) => {
+      const eventEntry = this.createEvent(row, mapping)
+      body.push(eventEntry)
+    })
+
+    winston.debug("BODY", JSON.stringify(body))
```
phillipperalez: nit: remove this and the debug comment

phillipperalez:
https://github.com/4mile/looker-actions/blob/658e55ca29f8a1df6917087620f9819823185e16/src/actions/mparticle/mparticle_transaction.ts#L227

Since a table_calculation isn't in the sample_data that you made, you are trying to do a `foreach` on something that is null

src/actions/mparticle/mparticle_transaction.ts
```diff
@@ -0,0 +1,287 @@
...
+      const err = "Each row must specify at least 1 identity tag."
+      this.errors.push(err)
+      throw err
+    }
+
+    return mapping
+  }
+
+  protected getTag(field: ExploreField): string {
+    // tslint:disable-next-line: strict-type-predicates
```
jj57347: The linter _really_ thinks these are always falsy / truthy (they're not), so I had to split them up and disable the strict type predicates here. I'm open to other suggestions to get it to work without disabling the linter, but I believe it's in error here

phillipperalez:
```
  protected getTag(field: ExploreField): string {
    // tslint:disable-next-line: strict-type-predicates
    if (field.tags === undefined || field.tags[0] === undefined) {
      return ""
    }
    const tag = field.tags.find((t) => t.startsWith("mp_"))
    return tag ? tag : ""
  }
```
at least minimizes the lint disabling

PR #299, opened 2019-11-28
Kloudio Action to push data to Google Sheets

Aniruddh-Kloudio:
This pull request contains the code to push data from Looker to Google Sheets using the Kloudio Action.


src/actions/kloudio/kloudio.ts
```diff
@@ -0,0 +1,177 @@
...
+    if (!( request.formParams.url)) {
+        throw "Missing Google sheets URL"
+    }
+
+    const qr = request.attachment.dataJSON
+    if (!qr.fields || !qr.data) {
+      throw "Request payload is an invalid format."
+    }
+
+    winston.info(typeof request.attachment.dataJSON)
```
phillipperalez: make this a debug statement or remove

src/actions/kloudio/kloudio.ts
```diff
@@ -0,0 +1,177 @@
...
+    try {
+        const newUri = API_URL.replace(/['"]+/g, "")
+        const lambdaResponse = await https.post({
+        url: newUri,
+        headers: {"Content-Type": "application/json"},
+        json: true,
+        body: data,
+         }).catch((_err) => {
+           const error = JSON.parse(_err)
+           winston.info("parsing error code" + error.emailCode)
```
phillipperalez: should this be an error level message? Also can this potentially log PII?

src/actions/kloudio/kloudio.ts
```diff
@@ -0,0 +1,177 @@
...
+          })
+
+        if (!lambdaResponse.success || lambdaResponse.success === false) {
+          winston.info("lambda url resp is not sucess " + lambdaResponse)
+          response = { success: false, message: lambdaResponse.message }
+        } else {
+          response = { success: true, message: lambdaResponse.message }
+        }
+    } catch (e) {
+      winston.info("Inside catch statement" + e)
```
phillipperalez: should this be an error message?

src/actions/kloudio/test_kloudio.ts
```diff
@@ -0,0 +1,44 @@
...
+        }, {
+          label: "Google Sheets URL",
+          name: "url",
+          required: true,
+          type: "string",
+        }],
+      }).and.notify(done)
+    })
+  })
+})
```
phillipperalez: We need tests to at least validate what happens when invalid urls or params or used to see what sort of errors and responses are passed back. I don't see any tests for the execute path

src/actions/kloudio/kloudio.ts
Aniruddh-Kloudio: Yes this has been made into an error level message. This does not log PII like user credentials or any other data. It logs generic error status like "GENERIC_ERROR" or "INVALID_PARAM"

src/actions/kloudio/test_kloudio.ts
Aniruddh-Kloudio: I have added a couple of unit test cases to check the cases when invalid api key or urls are entered.

Aniruddh-Kloudio:
Made the changes. Please let me know in case more are needed.

package.json
```diff
@@ -85,7 +85,9 @@
     "twilio": "^3.8.0",
     "typescript": "^3.0.1",
     "uuid": "^3.1.0",
-    "winston": "^2.3.1"
+    "winston": "^2.3.1",
+    "yarn": "^1.19.0",
```
phillipperalez: why is yarn added here?

Aniruddh-Kloudio: This was added by mistake...removed it.

src/actions/kloudio/kloudio.ts
```diff
@@ -0,0 +1,178 @@
+import * as https from "request-promise-native"
+
+import * as uuid from "uuid"
+import * as winston from "winston"
+import * as Hub from "../../hub"
+
+const signedUrl = "https://api.kloud.io/v1/tools/signed-url-put-object?key="
+const API_URL = "https://ecuoyevlzf.execute-api.us-west-2.amazonaws.com/prod/dest/send"
```
phillipperalez: Is this fine to expose on an open source project? If this is fine with you I'll merge it but maybe a comment with who Looker should email if someone makes a PR to change this would be a good safety measure

phillipperalez: That or you can give us a current url to deploy in an environment variable

phillipperalez: In fact I would prefer if you made this an environment variable that we could load on actionhub startup.

Aniruddh-Kloudio: Environment variables work for us if they can only be accessed by only our code & Looker's code. If it can be accessed by other people's code then the current setup works fine for us security wise. You can send an email to developer@kloud.io when someone raises a PR to change them. This is a monitored email.

phillipperalez: Let's make this an environment variable that we control, like `KLOUDIO_API_URL`. We will keep this in a secure store that is only read at startup of that server. If it needs to be update KloudIO can contact Looker to get it updated

phillipperalez: It will not be accessed by any other actions

Aniruddh-Kloudio: Ok I'll make the change and update the PR. How should the the values of these variables sent? with this PR or through an email?

src/actions/kloudio/test_kloudio.ts
Aniruddh-Kloudio: Hi @phillipperalez! I just wanted to ask whether I need to add any other test cases or the ones currently present are ok?

phillipperalez:
The current state of the icons getting sent to Helltool is by the image icons getting turned into base64 and saved in Looker. 643K is too large, try something much smaller, like less than 256x256 would be ideal

src/actions/kloudio/kloudio.ts
Aniruddh-Kloudio: Hi! Just to clarify, both the variables(signedUrl, API_URL) need to be replaced by environment variables right? or just the API_URL?

src/actions/kloudio/kloudio.ts
```diff
@@ -0,0 +1,178 @@
...
+
+async function uploadToS32(url: any, s3Data1: any) {
+
+  const santUrl = url.replace(/['"]+/g, "")
+  const response = https.put({
+    url: santUrl,
+    headers: {"Content-Type": "application/json"},
+    json: true,
+    body: s3Data1,
+     }).catch((_err) => { winston.error(_err.toString()) })
```
phillipperalez: `_err` is used so should it be `err` instead? the `_` is for parameters that aren't getting used

src/actions/kloudio/kloudio.ts
```diff
@@ -4,7 +4,7 @@ import * as uuid from "uuid"
 import * as winston from "winston"
 import * as Hub from "../../hub"
 
-const signedUrl: any = process.env.KLOUDIO_SIGNED_URL
+const signedUrl = "https://api.kloud.io/v1/tools/signed-url-put-object?key="
```
phillipperalez: why was this reverted back?

src/actions/kloudio/kloudio.ts
Aniruddh-Kloudio: ok this has been changed to "err".

src/actions/kloudio/kloudio.ts
Aniruddh-Kloudio: Initially I had thought both the variables had to be changed but since the previous PR comments & Shohei's email only mentioned "API_URL" I reverted it back since it was not asked.

snarron: Hey Anirrudh, my apologies. I think @phillipperalez was asking for both - basically not hard-coding URLs in general for these two variables.

Aniruddh-Kloudio: Hi Shohei! Thanks for the clarification. I will make the changes and update the PR.

Aniruddh-Kloudio: @snarron @phillipperalez I have made the changes by reading both values from env variables. The test file is using a test url not the live url.

PR #313, opened 2020-01-24
Remove marketo custom queue

phillipperalez:


src/actions/marketo/marketo_transaction.ts
```diff
@@ -70,15 +71,23 @@ export class MarketoTransaction {
...
+      return new Promise<void>((resolve, reject) => {
+        const chunk = rows.slice(0)
+        rows = []
+        this.processChunk(chunk).then((response: any) => {
+          results.push(response)
+          resolve()
+        })
+        .catch((error: any) => {
+          errors.push(error)
+          reject()
```
brln-looker: reject(error)

src/actions/marketo/marketo_transaction.ts
```diff
@@ -70,15 +71,23 @@ export class MarketoTransaction {
...
+    const results: Result[] = []
+    const errors: Result[] = []
+
+    const sendChunk = async () => {
+      return new Promise<void>((resolve, reject) => {
+        const chunk = rows.slice(0)
+        rows = []
+        this.processChunk(chunk).then((response: any) => {
+          results.push(response)
+          resolve()
```
brln-looker: probably `resolve(response)`? You might not be using it here but that's what I'd expect from the interface.

phillipperalez: This is fine because it is a void Promise

PR #324, opened 2020-02-05
Added default S3 region

phillipperalez:
Added default regions for S3. If getBucketLocation(...) for AWS S3 returns a null LocationConstraint, the default is "us-east-1"


src/actions/sagemaker/sagemaker_train_linearlearner.ts
```diff
@@ -376,8 +376,11 @@ export class SageMakerTrainLinearLearnerAction extends Hub.Action {
       Bucket: bucket,
     }
     const response = await s3.getBucketLocation(params).promise()
-
-    return response.LocationConstraint
+    if (response.LocationConstraint) {
+      return response.LocationConstraint
+    } else {
+      return "us-east-1"
```
annguy3n: nit: maybe add a const for this

Issue #227, opened 2018-12-14
[slack] Cannot see private channels


ravimody:
I added the bot to a private channel but it doesn't give an option to send results to that channel. It can successfully send to any user and to public channels its added to. I gave it these three permissions:

- channels:read
- users:read
- files:write:user

Do I need to give it more permissions to enable sends to private channels?

gwaldmantoast:
+1

sumeshnmenon:
+1

wilg:
Looks like this is not currently supported. The relevant permission is [`groups:read`](https://api.slack.com/scopes/groups:read), but currently the Slack action does not attempt to list groups.

We definitely want to address this. Thanks for the reports!

mghita-guardian:
+1

PR #327, opened 2020-02-18
allow legacy slack action to post to private channels

jakeo:
This adds private channel support to the legacy slack action. Additionally, it enables private channel support to customer-hosted instances via the legacy slack action.

This undoes most of the changes in https://github.com/looker/actions/pull/279 which introduced this due to token confusion. `xoxp` vs `xoxb` token usage can be handled in documentation.

fixes: https://github.com/looker/actions/issues/227


phillipperalez:
Thank you for the change!

PR #332, opened 2020-02-27
Add support for Shared Drives for Google Sheets/Drive

phillipperalez:
Adds proper corpus values to Google Drive/Sheets and the ability to select what Drive to send the data.
Also adds ability to select top level Drive Root as selectable folder

### DIFF
--- a/src/actions/google/drive/google_drive.ts
+++ b/src/actions/google/drive/google_drive.ts
@@ -108,7 +108,6 @@ export class GoogleDriveAction extends Hub.OAuthAction {
             return mergedFiles
           }
           const paginatedFiles = await pagedFileList([], await drive.files.list(options))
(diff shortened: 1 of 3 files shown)


src/actions/google/drive/google_drive.ts
```diff
@@ -85,30 +108,34 @@ export class GoogleDriveAction extends Hub.OAuthAction {
             return mergedFiles
           }
           const paginatedFiles = await pagedFileList([], await drive.files.list(options))
+          winston.info(JSON.stringify(paginatedFiles))
```
brln-looker: do you want this in prod?

phillipperalez: we do not, thanks and good catch

PR #333, opened 2020-02-27
Add parameters for Shared Drive uploads

phillipperalez:
Adds ability to upload data to Shared Drives correctly

### DIFF
--- a/src/actions/google/drive/google_drive.ts
+++ b/src/actions/google/drive/google_drive.ts
@@ -201,12 +201,23 @@ export class GoogleDriveAction extends Hub.OAuthAction {
        mimeType: this.mimeType,
        parents: request.formParams.folder ? [request.formParams.folder] : undefined,
      }
+
      return request.stream(async (readable) => {
-       return drive.files.create({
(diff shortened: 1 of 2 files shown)


jdavidson:
I'm not super familiar with the API options but this looks good to me

src/actions/google/drive/google_drive.ts
```diff
@@ -201,12 +201,23 @@ export class GoogleDriveAction extends Hub.OAuthAction {
...
          },
+       }
+
+       if (request.formParams.drive !== undefined && request.formParams.drive !== "mydrive") {
+         driveParams.requestBody!.driveId! = request.formParams.drive
+         driveParams.supportsAllDrives = true
+       }
+
+       return drive.files.create(driveParams).catch((e) => {
+         winston.error(JSON.stringify(e.errors))
```
jdavidson: any risk writing the full error?

phillipperalez: It should be fine but I will demote to debug to be safe

PR #343, opened 2020-05-07
Infer file extension for streaming formats

rainiera:
This change proposes inferring the file extension for webhook payloads that don't have the `attachment` field (equivalent to streaming formats).

One use-case this solves is when customers attempt to send a query result as an Excel file to Slack. The Slack API is dependent on an `filename` field to determine file type, and absent this field, what gets posted is a file called `Untitled` that shows up as a `zip` instead of a `xlsx` file.

### DIFF
--- a/src/actions/sftp/sftp.ts
+++ b/src/actions/sftp/sftp.ts
@@ -32,7 +32,7 @@ export class SFTPAction extends Hub.Action {
         throw "Needs a valid SFTP address."
       }
       const data = request.attachment.dataBuffer
-      const fileName = request.formParams.filename || request.suggestedFilename() as string
+      const fileName = request.formParams.filename || request.suggestedFilename()
       const remotePath = Path.join(parsedUrl.pathname, fileName)
 
       client.put(data, remotePath)
--- a/src/hub/action_request.ts
+++ b/src/hub/action_request.ts
@@ -356,11 +356,11 @@ export class ActionRequest {
         return sanitizeFilename(`looker_file_${Date.now()}.${this.attachment.fileExtension}`)
       }
     } else {
-      if (this.params.format) {
+      if (this.formParams.format) {
         if (this.scheduledPlan && this.scheduledPlan.title) {
-          return sanitizeFilename(`${this.scheduledPlan.title}.${formatToFileExtension(this.params.format)}`)
+          return sanitizeFilename(`${this.scheduledPlan.title}.${formatToFileExtension(this.formParams.format)}`)
         } else {
-          return sanitizeFilename(`looker_file_${Date.now()}.${formatToFileExtension(this.params.format)}`)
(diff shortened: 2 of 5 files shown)


src/hub/action_request.ts
```diff
@@ -355,6 +355,14 @@ export class ActionRequest {
       } else {
         return sanitizeFilename(`looker_file_${Date.now()}.${this.attachment.fileExtension}`)
       }
+    } else {
```
JimRottinger, line 358: Let's add some tests to https://github.com/looker/actions/blob/master/test/test_action_request.ts for the suggested filename. Lots of conditionals and now a util function in here now.

src/hub/action_request.ts
```diff
@@ -355,6 +355,14 @@ export class ActionRequest {
...
+      if (this.formParams.format) {
```
JimRottinger, line 359: Not necessary but I would recommend changing this to an `else if` and always having a return in every codepath.

```js
if (this.attachment) {
  /** Do sstuff */
}
else if (this.formParams.format) {
  /** do else if stuff */
}

return 'defaultSuggestedFilename'
```

src/actions/sftp/sftp.ts
```diff
@@ -32,7 +32,7 @@ export class SFTPAction extends Hub.Action {
         throw "Needs a valid SFTP address."
       }
       const data = request.attachment.dataBuffer
-      const fileName = request.formParams.filename || request.suggestedFilename() as string
+      const fileName = request.formParams.filename || request.suggestedFilename()
```
rainiera, line 35: since changes to `request.suggestedFilename` mean it no longer has a chance at being `undefined`, the linter said type assertion here is no longer necessary.

src/hub/action_request.ts
```diff
@@ -355,7 +355,15 @@ export class ActionRequest {
...
       }
+    } else if (this.formParams.format) {
+      if (this.scheduledPlan && this.scheduledPlan.title) {
+        return sanitizeFilename(`${this.scheduledPlan.title}.${formatToFileExtension(this.formParams.format)}`)
+      } else {
+        return sanitizeFilename(`looker_file_${Date.now()}.${formatToFileExtension(this.formParams.format)}`)
+      }
     }
+    winston.warn("Couldn't infer file extension from action request, using default filename scheme")
+    return sanitizeFilename(`looker_file_${Date.now()}`)
```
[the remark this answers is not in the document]
annguy3n: I don't think AH support whitelabel (@phillipperalez, correct me if i'm wrong here). They can host AH themselves as well, and change the prefix. I'd say it's out of the scope of this since it's the same as L356: https://github.com/looker/actions/pull/343/files#diff-69063cbe99915fa7386cdacf954948c5L356

PR #352, opened 2020-05-20
Make sure quotes properly formatted for CSV lines

phillipperalez:
Before this change a record like `This tricky",string` would be sent incorrectly as `"This tricky",string"` which is invalid CSV. This PR reformat all `"` in strings to be correct `""` - making the example `This tricky"",string` and correct CSV

### DIFF
--- a/src/actions/google/drive/sheets/google_sheets.ts
+++ b/src/actions/google/drive/sheets/google_sheets.ts
@@ -173,6 +173,10 @@ export class GoogleSheetsAction extends GoogleDriveAction {
                             reject(e)
                         })
                     }
+                    const lineData = line.map((v: any) => {
+                        v = (v as string).replace(/\"/g, "\"\"")
+                        return `"${v}"`
+                    }).join(",") as string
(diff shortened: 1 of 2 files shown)


src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -173,6 +173,10 @@ export class GoogleSheetsAction extends GoogleDriveAction {
                             reject(e)
                         })
                     }
+                    const lineData = line.map((v: any) => {
+                        v = (v as string).replace(/\"/g, "\"\"")
```
delaneyparker, line 177: This might be relevant. tl;dr - maybe use tr() instead?
https://blog.appsignal.com/2019/08/20/clean-up-strings.html

delaneyparker: Nevermind, for JS this looks to be as fast as we can get...
https://stackoverflow.com/questions/9750338/dynamic-vs-inline-regexp-performance-in-javascript

PR #355, opened 2020-05-21
Move resize logic until after request size is locked

phillipperalez:
Lock request size before resizing sheet, reducing error rates that the batch update will mis-size the sheet


phillipperalez:
CI:TEST

PR #358, opened 2020-06-27
🔥 [New Actions]: Hubspot Companies and Contacts

luskin:
This PR introduces two new Looker Actions:

- Hubspot Contacts
- Hubspot Companies. 

**Usage**
Looker users can now apply a `hubspot_contact_id` or `hubspot_company_id` dimension tag and use this action to batch update their contact or company properties. We are using the new [Hubspot Typescript NodeJS API V3 SDK](https://github.com/HubSpot/hubspot-api-nodejs) which by default includes throttling and retries, making sure you won't get hit with those pesky 10 call per second hubspot limits!

**Included**
- Handling of hidden fields
- Passing your Hubspot API key in as a parameter during action setup
- Validation to ensure that if you are using the Contacts action, you **MUST** have a `hubspot_contact_id` tagged column and similarly for the Companies action.
- The abstraction of the core Hubspot logic so that future additions of Deals or other Hubspot CRM products can be easily implemented.
- Batching calls by 100 as to stay under the HubSpot batch update limit
- A handful of tests to make sure there are no regressions on core functionality

Screen Shot 2020-06-27 at 12 04 04 PM

Screen Shot 2020-06-27 at 12 04 44 PM

Screen Shot 2020-06-27 at 12 05 00 PM


src/actions/hubspot/hubspot.ts
```diff
@@ -0,0 +1,257 @@
...
+  label: string
+  description: string
+  call: HubspotCalls
+  tag: HubspotTags
+}
+
+async function delay(ms: number) {
+  // tslint continually complains about this function, not sure why
+  // tslint:disable-next-line
+  return new Promise((resolve) => setTimeout(resolve, ms))
```
phillipperalez, line 30: it's probably mad about not having a reject or .catch function

src/actions/hubspot/hubspot.ts
```diff
@@ -0,0 +1,257 @@
...
+      sensitive: false,
+    },
+  ]
+  supportedActionTypes = [Hub.ActionType.Query]
+  usesStreaming = true
+  supportedFormattings = [Hub.ActionFormatting.Unformatted]
+  supportedVisualizationFormattings = [
+    Hub.ActionVisualizationFormatting.Noapply,
+  ]
+  executeInOwnProcess = true
```
phillipperalez, line 68: I am glad you did this already. Out of curiosity what is the biggest dataset you have tried streaming through this action and how big do you expect a "normal" request to be?

src/actions/hubspot/hubspot.ts
```diff
@@ -0,0 +1,257 @@
...
+    const errors: Error[] = []
+
+    try {
+      await request.streamJsonDetail({
+        onFields: (fields) => {
+          fieldset = Hub.allFields(fields)
+          hubspotIdFieldName = this.getHubspotIdFieldName(fieldset)
+          if (!hubspotIdFieldName) {
+            const error = `Dimension with the ${this.tag} tag is required`
+            winston.error(error)
```
phillipperalez, line 170: if you can add the webhookID to the error and log messages it makes it a lot easier to debug issues when they pop up

phillipperalez:
Also @luskin we just need to make sure there is a license file in this PR, like the one below

https://github.com/looker/actions/blob/master/src/actions/auger/LICENSE

src/actions/hubspot/hubspot.ts
luskin: @phillipperalez depends entirely on the size of the company, for us we're passing in a few thousand at a time, upwards of maybe 5k rows from Looker.

src/actions/hubspot/hubspot.ts
luskin: Yeah, I tried that but still receiving: `Explicit type parameter needs to be provided to the constructor`

phillipperalez:
@luskin if we could plan a demo at some point that would be great too. I am excited to see what you have built.

luskin:
@phillipperalez happy to do that, how would you like to set that up?

phillipperalez:
What is the best way to contact you?

phillipperalez:
Actually go ahead and email blocks@looker.com and we will organize through email

phillipperalez (approved):

PR #360, opened 2020-07-14
Phillip/unite slack calls

phillipperalez:
convert legacy Slack actions to use `conversations.list` instead of the deprecated `channels.list` API call

https://api.slack.com/methods/channels.list


annguy3n:
Can you update the readme to add groups read permission? i think it's required if the tokens aren't bot tokens.

phillipperalez:
No longer valid

PR #363, opened 2020-07-29
Phillip/fix custom action hub

phillipperalez:
This change will fix to the action hub repo "looker/actions" to correctly build with a "yarn install"


src/hub/action_request.ts
```diff
@@ -1,12 +1,13 @@
 import * as express from "express"
 import * as oboe from "oboe"
 import * as httpRequest from "request"
-import * as sanitizeFilename from "sanitize-filename"
 import * as semver from "semver"
 import { PassThrough, Readable } from "stream"
 import * as winston from "winston"
 import { formatToFileExtension, truncateString } from "./utils"
 
+let sanitizeFilename = require("sanitize-filename")
```
brln-looker, line 9: can you do this as an import for consistency?

phillipperalez: When this particular library updated it just stopped working in the `import` style unfortunately. This was a known issue that they were fine with. It was causing ECMAinterop issues when used in the different styles of `import` I could think of

brln-looker: you tried `import * as sanitize-filename`?

phillipperalez: yes and that shows up with the interop error

PR #369, opened 2020-09-04
Pagerduty service integration spike

jgogstad:
* REST API integration works
* Event API integration works

The following user story is green:

> As a developer
> I can create a scheduled query in Looker than will generate alerts in PagerDuty

Remaining on epic:

* Templating messages to PagerDuty from Looker resultset
* Surface more fields from PagerDuty API in Looker
* Implement auto resolving of PD incidents when resultset changes
* Tests
* Deploy to CloudRun


src/actions/pagerduty/pagerduty.ts
```diff
@@ -0,0 +1,110 @@
...
+    sensitive: true,
+  }]
+  usesStreaming = false
+
+  async execute(request: Hub.ActionRequest): Promise<Hub.ActionResponse> {
+    function validationError(e: Hub.ValidationError): Hub.ActionResponse {
+      return new Hub.ActionResponse({ message: 'Validation error', success: false, validationErrors: [e] })
+    }
+
+    const form = PagerDutyForm.create(request.formParams as unknown)
```
jgogstad, line 29: This is weird, let's revisit. `formParams` is of type `ParamMap`:

```
export interface ParamMap {
  [name: string]: string | undefined
}
```

src/actions/pagerduty/pagerduty.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (form == null)
+      return new Hub.ActionResponse({ message: 'Invalid form specification, please contact administrator', success: false, })
+
+    const apiKey = this.getApiKey(request)
+    const resultSet = this.parseResultSet(request.attachment)
+
+    if (typeof apiKey !== "string")
+      return validationError(apiKey)
```
jgogstad, line 37: `apiKey` is of type `ValidationError | string`. `ValidationError` is an interface. This conditional would read better if something like this compiled:

```
if (apiKey instanceOf ValidationError) return validationError(apiKey)
```

src/actions/pagerduty/pagerduty_form.ts
```diff
@@ -0,0 +1,100 @@
...
+            label: 'Maximum number of incidents',
+            name: PagerDutyForm.MaxIncidentsKey,
+            description: 'Maximum number of incidents to create based on result set',
+            type: 'string',
+            default: '10',
+            required: true
+        }
+    ]
+
+    static create(json: any): PagerDutyForm | undefined {
```
jgogstad, line 43: Is this how you create disjunctions with the unit type? I.e. is there a `Option<PagerDutyForm>` or something similar that is more idiomatic?

src/actions/pagerduty/pagerduty_form.ts
```diff
@@ -0,0 +1,100 @@
...
+        ]
+
+        const ok = keys.reduce((p: boolean, key: string) => p && json[key] != null, true)
+
+        if (!ok) return undefined
+
+        return new PagerDutyForm(json[PagerDutyForm.ServiceKeyKey], json[PagerDutyForm.DeDeuplicationIdColumnKey],
+            json[PagerDutyForm.AutoResolveKey],
+            json[PagerDutyForm.MaxIncidentsKey]
+        )
```
jgogstad, line 61: I'm impressed this compiles. Is there a better/more-idiomatic way to go from "any" to "Maybe PagerDutyForm"?

jgogstad:
Opened on wrong repo, never mind

PR #374, opened 2020-10-28
Adding Google Cloud AutoML data action

samudzi:
This action allows the user to send the results of an explore or look to GC AutoML as a new AutoML table.  It additionally creates a csv file within the user-specified GCS bucket.

Limitations:
-user needs to manually change the field labels for each field in the visualization to lowercase.  only alphanumeric + underscores
-gcs bucket and automl dataset must be in the same region

This also fixes an issue I found with the private key input in the other google data actions.  Apparently, a number of clients are using them, so did not want to break anything by including my fixes for those two other actions.


phillipperalez:
You need to resolve the issue with the PR that it is literally deleting almost every file that isn't the changes first before I can review this

PR #378, opened 2020-12-24
Add Teams Action

KumakuraKoki:
This Action allows a looker user to send messages to teams via incoming webhook


src/actions/teams/teams.ts
```diff
@@ -0,0 +1,138 @@
...
+  supportedFormattings = [Hub.ActionFormatting.Unformatted]
+  supportedVisualizationFormattings = [
+    Hub.ActionVisualizationFormatting.Noapply,
+  ]
+  params = []
+
+  async execute(req: Hub.ActionRequest) {
+    let response = { success: true, message: "success" }
+
+    if (!req.formParams.webhookUrl) {
```
jdavidson, line 21: will the webhook for teams be changing on a frequent basis and will business users have knowledge of this endpoint? if not, it'd be good to make the webhook part of the action set up for example like this: https://github.com/looker/actions/blob/82767e2170b11ec53bac4c37e9a16e3a43906fa4/src/actions/marketo/marketo.ts#L10

the error should pass back to Looker in an Action response e.g. https://github.com/looker/actions/blob/82767e2170b11ec53bac4c37e9a16e3a43906fa4/src/actions/marketo/marketo_transaction.ts#L139

KumakuraKoki:
@jdavidson 
Incoming Webhooks can be configured for each Teams' channel and not only for administrators but also for business users can check the endpoint. if embedded in  `param` , the action will be delivered to only one Teams' channel, and I thought it is too small a destination and too small to deliver to.

@theo-browne
I thought that if I don't specify any format in the source code, all the format choices will be shown to the user by default (or not?).
In this case, since I wanted to send only the link of the Looker to Teams, I didn't want the user to be aware of the format choices, and if only one format is selected, I thought the specification was that the user would not see the choices, so I just selected one.

theo-browne (approved):
LGTM @phillipperalez tagging you in case you want to take a look

PR #380, opened 2021-01-05
Replace dot with underscore in property names for Hubspot action

mvaerle:
We've been trying to use the Hubspot action to send some data from looker into Hubspot company properties. This is not working due to the fact that looker sends data in the form of `view.dimension`, for example: `customer.revenue` to Hubspot properties. However Hubspot properties do not support the use of dots `.` in their names.

This PR replaces those dots with underscores, which is the default behaviour when adding a dot in a property name in the Hubspot UI.

Tagging @luskin since you were the original author, thanks for creating this custom action!

### DIFF
--- a/src/actions/hubspot/hubspot.ts
+++ b/src/actions/hubspot/hubspot.ts
@@ -181,8 +181,8 @@ export class HubspotAction extends Hub.Action {
                 fieldName !== hubspotIdFieldName &&
                 hiddenFields.indexOf(fieldName) === -1
               ) {
-                const safeFieldName = fieldName.replace(".", "_");
-                properties[safeFieldName] = fieldSet.value;
+                const safeFieldName = fieldName.replace(".", "_")
+                properties[safeFieldName] = fieldSet.value
               }
             })
 
--- a/src/actions/hubspot/hubspot.ts
+++ b/src/actions/hubspot/hubspot.ts
@@ -181,7 +181,8 @@ export class HubspotAction extends Hub.Action {
                 fieldName !== hubspotIdFieldName &&
                 hiddenFields.indexOf(fieldName) === -1
               ) {
-                properties[fieldName] = fieldSet.value
+                const safeFieldName = fieldName.replace(".", "_");
+                properties[safeFieldName] = fieldSet.value;
               }
             })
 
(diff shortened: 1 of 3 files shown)


phillipperalez:
I'm going to check with the original writers about the effects this could have.  Definitely going to check with @luskin to make sure that this change is good. My main concern that needs to be checked out is that if people have scheduled Hubspot actions, that this will cause the action to act differently and cause problems.

That isn't to say I don't think the change is a bad idea, it seems good. Just a problem in how much communication I need to do before I can approve and merge this into a production build. Unfortunately one of those "the bug is currently a feature" issues.

mvaerle:
I understand to check with the original author, no problem! 

Some extra context below:

Theoretically this will cause existing actions to behave differently, since any action that's sending a dimension with a dot in it will now send it under a different property name. However since it's not possible to create a property with a dot in it on hubspot this should not cause any existing implementations to break, since they must've been already broken.

I can't reference to existing hubspot documentation, since the allowed characters are not documented as far as I could find. However this reply to a forum post by a hubspot employee lists the allowed input for a property name:

https://community.hubspot.com/t5/APIs-Integrations/Is-there-any-limitation-on-the-name-of-the-properties-of/td-p/232575#:~:text=Property%20names%20must%20contain%20only,must%20start%20with%20a%20letter.

Also I've figured a workaround for this problem that probably everyone using this action must be using, but that I couldn't find anywhere. You can create a custom dimension and looker will send that name without prepending the view to it. For example: `view_one.property_one` can have a custom dimension called `view_one_property_one` and that will work. It's a bit of a drag to create custom dimensions for everything that this integration can use, but it works ;-)

luskin:
@mvaerle You make a good point here. When I originally wrote this we were only using custom dimensions so I did not consider the case when a default dimension would be passed in from Looker. I don't see any glaring issues with this, other than the `.replace` call itself. I would say out of an abundance of caution we should probably safeguard against any future where there might be multiple `.`'s in a property name.  A simple change to: `fieldName.replace(/./g, '_')` I _think_ should resolve it (please make sure my syntax is correct there).

mvaerle:
@luskin Good catch, I agree and have updated the code according to your suggestion with a minor syntax change

phillipperalez (approved):
If @luskin doesn't have any other feedback this should be fine. I'll ping this thread after merging to let people know when this hits our prod deployment

LookingAbercrombie:
Any chance we can get a patch update for this? I have a customer running into this issue and just need some guidance on if I need to tell them March with the 21.4 release or if they can get it sooner. 

Thank you!

aryeh-looker:
Merging on behalf of @phillipperalez as he's out at the moment. Looks good, we'll work on getting this into prod. Thanks @mvaerle and @luskin!

PR #383, opened 2021-02-01
Add "csv_zip" in supported_formats, which is available from integration

yingbiao:
`csv_zip` option is available in `api_types/integration.ts`, but is not in documenation


phillipperalez (approved):
LGTM, thank you!

PR #385, opened 2021-02-02
Add Google Analytics and Google Ads actions

vpoiesz:
The time has come! 🎉 

This branch includes both the **Google Analytics Data Import** action _and_ the **Google Ads Customer Match** action. (They share common code and it was easier to handle them together...)

The **Google Analytics Data Import** action feels production ready and has already been beta tested by customers. It includes what should be a complete suite of automated integration tests. There is a basic README which will be converted into a Community post that we can later link to instead.

The **Google Ads Customer Match** action is feature complete, but still in the alpha phase. The test suite and documentation for it still need more work.

There are **two new top-level dependencies:**

- lodash - "A modern JavaScript utility library delivering modularity, performance & extras"
- gaxios - "An HTTP request client that provides an axios like interface over top of node-fetch." [This is a Google-owned packaged already used by the googleapis package under the hood.]

Note that the full test suite does **not** run clean for me:

```
  1) QueueTestAction unit tests for ProcessQueue
       action
         ExecuteProcessQueue returns success on successful response:
     AssertionError: expected promise to be fulfilled but it was rejected with 'AssertionError: expected false to equal true'
  

  2) QueueTestAction unit tests for ProcessQueue
       action
         ExtendedProcessQueue returns success on successful response:
     AssertionError: expected promise to be fulfilled but it was rejected with 'AssertionError: expected false to equal true'
 ````

Although I don't think my code is contributing to this. Also, several of the Marketo action tests are exceeding the default timeout for async tests.

I'm looking forward to any and all feedback!

### CODE REVIEW

**Comment by vpoiesz** [comment]
@phillipperalez I've left this as a draft PR. We probably don't want to merge directly to master, but I don't believe I have the ability to create new remote branches in this repo. If you create a new branch (or give me the permissions) we can update the merge target.

File: src/actions/google/ads/lib/Customer_Match_Test_File_-_Contact_Information.csv [hunk 1]
```diff
@@ -0,0 +1,52 @@
+# Instructions:
```
vpoiesz, line 1: This file is provided by Google as an example of the data normalization & hash procedure. I added it to the code to help serve as documentation. I also hope to use it directly within the automated test suite to validate our logic.

src/actions/google/analytics/lib/ga_worker.ts
```diff
@@ -0,0 +1,352 @@
...
+      throw new MissingAuthError("User state was missing or did not contain tokens & redirect")
+    }
+
+    this.userState = tmpState
+    this.formParams = hubRequest.formParams
+    this.gaClient = this.makeGAClient()
+  }
+
+  log(level: string, ...rest: any[]) {
+    return this.actionInstance.log(level, ...rest)
```
phillipperalez, line 41: This log is not valid in actionhub if it is not using Winston

vpoiesz: Hey @phillipperalez , I'm not sure I follow. This method redirects to the action instance which in turn uses Winston. Are you saying that it isn't guaranteed Winston will be available, and that actions shouldn't call it directly? I didn't see any other logging interfaces, but could be missing something. I didn't use Winston directly in all the helper code because I wanted to be able to encapsulate some logic like adding the prefix.

vpoiesz:
After discussing with Aryeh, I am closing this PR in favor of a new one that splits up the two actions.

PR #388, opened 2021-02-22
Theo/414

theo-browne:
Addresses an issue where the wrong sheet is cleared out when the first sheet is hidden

### DIFF
--- a/src/actions/google/drive/sheets/test_google_sheets.ts
+++ b/src/actions/google/drive/sheets/test_google_sheets.ts
@@ -143,9 +143,21 @@ describe(`${action.constructor.name} unit tests`, () => {
                   ],
                 },
               }),
-              values: {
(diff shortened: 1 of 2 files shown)


src/actions/google/drive/sheets/test_google_sheets.ts
```diff
@@ -320,9 +356,21 @@ describe(`${action.constructor.name} unit tests`, () => {
                   ],
                 },
               }),
-              values: {
-                clear: async () => Promise.resolve(),
-              },
+              batchUpdate: async () => Promise.resolve({
+                spreadsheetId: "1",
```
annguy3n, line 360: nit: maybe have a mock var to reuse it in these tests

annguy3n (approved):

PR #392, opened 2021-03-02
Ensure dotenv is configured before other imports

vpoiesz:
I noticed this in the course of unrelated development. My new actions issue a warning to the log when they fail to register because of missing env vars. I noticed these warnings were being printed during the ProcessQueue tests. So I took a look and saw that the action files were being imported before calling `dotenv.config()`. Which means that if any actions were depending on certain env vars at import time (like mine) they would fail when executed in a separate process.

Instead I switched the imports to `import "dotenv/config"` which is the recommendation that plays best with our existing setup: https://www.npmjs.com/package/dotenv#how-do-i-use-dotenv-with-import-

### DIFF
--- a/bin/generate_key.ts
+++ b/bin/generate_key.ts
@@ -1,8 +1,7 @@
-import * as crypto from "crypto"
-import * as dotenv from "dotenv"
-import * as keys from "../src/server/api_key"
+import "dotenv/config"
 
-dotenv.config()
+import * as crypto from "crypto"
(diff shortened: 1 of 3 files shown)


vpoiesz:
Hey @phillipperalez , I think this PR is worth considering!

If we wanted to run the new Google actions in the XPC then they actually wouldn't work, because the new process will try to register the actions before configuring dotenv. The actions won't register themselves with the Hub if they detect that the necessary env vars are missing.

PR #395, opened 2021-03-25
New action: Google Analytics Data Import

vpoiesz:
This action allows Looker users to send report data to a "custom data set" within their Google Analytics accounts. Custom data sets can be used to enrich hit data within GA for more accurate reporting, audience segmentation, and audience remarketing. This is integration an excellent demonstration of Looker's data delivery capabilities and helps "close the loop" by taking advanced analysis done in BQ+Looker and feeding those insights back into the Google Marketing products.

The Analytics client library is provided by the "googleapis" npm package, which is already a direct project dependency because it is used by the Google Drive action too.

This action uses the AH oauth functionality and will thus require oauth_client_id and oauth_client_secret to be passed as environment variables. The action will not register itself with the hub if they are not present. I have already been engaged with the Looker Ops team to create the necessary GCP projects for the API gateway and oauth credentials. 

This PR includes several helper files in the `src/actions/google/common` folder. I have another action in development which will use these common services as well.

I have already successfully beta tested the action with several customers. It feels feature complete and ready for launch. There is a thorough automated test suite. I look forward to any and all feedback.


src/actions/google/analytics/data_import.ts
```diff
@@ -0,0 +1,145 @@
...
+        await gaWorker.deleteOtherFiles()
+        this.log("info", `${currentStep} completed.`)
+      }
+
+      // All is well if we made it this far
+      this.log("info", "Execution completed successfully.")
+      return wrappedResp.returnSuccess()
+    } catch (err) {
+      this.log("error", "Execution error:", err.toString())
+      this.log("error", "Exeuction errror JSON:", JSON.stringify(err))
```
vpoiesz, line 100: I need to check whether `JSON.stringify(err)` will inadvertently log the headers associated with the request, which would include the oauth access token. I originally added this because the google api error objects contain a bunch of additional properties with more helpful info than just the message (e.g. on a 400 Bad Request, it will include a list of the fields with problems).

vpoiesz:
Hey @phillipperalez / @aryeh-looker - any thoughts so far?

src/actions/google/analytics/data_import.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  async form(hubReq: Hub.ActionRequest) {
+    const wrappedResp = new WrappedResponse(Hub.ActionForm)
+
+    try {
+      const gaWorker = await GoogleAnalyticsActionWorker.fromHubRequest(hubReq, this)
+      wrappedResp.form = await gaWorker.makeForm()
+      this.log("info", "Form generation complete")
```
phillipperalez, line 112: unless there is a webhookId associated with these sorts of log lines, can we make this debug? Without the webhookID this is just noise in the logs

phillipperalez:
Other than that comment about webhookIDs, I think we are good to go here!

src/actions/google/analytics/data_import.ts
vpoiesz: Good point! I have refactored the logging system to include webhookId wherever possible, including this line. Although I'm still happy to change the levels / text further as needed.

vpoiesz:
Excellent! I've committed that fix plus the bit about sanitizing error objects (which were including the oauth secrets).

Also from an ops perspective I'm ready with the GCP project for staging creds. And our DevOps team is standing by to make the prod project + secrets whenever we are.

phillipperalez (approved):

vpoiesz:
@phillipperalez I don't have merge permissions - would you care to do the honors? 😃

PR #398, opened 2021-04-01
Add retries and increase batch limit for google sheets

theo-browne:


src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -9,7 +9,7 @@ import Drive = drive_v3.Drive
 import Sheet = sheets_v4.Sheets
 import {GoogleDriveAction} from "../google_drive"
 
-const MAX_REQUEST_BATCH = 4096
+const MAX_REQUEST_BATCH = 100
```
phillipperalez, line 12: should this be 4096 and not 100, I assume this is an oversight

theo-browne: You're correct, I decreased it when i was testing

phillipperalez (approved):

PR #401, opened 2021-05-05
[HEAP-22744] fix bug

SSYoung:
In the last PR i accidentally had a newer version of typescript running, which allowed me to use language syntax that doesn't exist in the TS version used in prod. Also, the import of `{ Field }` instead of using `Hub.Field` caused a compilation error


SSYoung:
pushed to the wrong fork

PR #404, opened 2021-05-12
New action: Google Ads Customer Match

vpoiesz:
Hi @phillipperalez , here is the PR we talked about.

I actually think this was ready to submit anyway... aside from the test suite, but I am already on the hook for any bug fixes 😉 

So if you are ok with the code, it's just a question of when we are ready to deploy to prod. Which from our end only needs the oauth verification to finish and put marketing materials in place. If we do want to hold off then after I record the demo I can submit another PR to turn off the registration for this action. Or, we could merge this PR to a non-master branch then deploy that to the dev server, since I noticed jenkins can take the branch name as a param. Totally up to you, since you have to trigger the merge anyway.

Happy to discuss further if needed. Thank you!


phillipperalez (approved):

vpoiesz:
Thank you so much, @phillipperalez ! This is huge!!

PR #415, opened 2021-06-24
Check to resize sheet before final flush

phillipperalez:
Check to resize sheet before doing the final flush to the sheet


phillipperalez:
@theo-browne can you take a quick look?

theo-browne (approved):
LGTM

PR #420, opened 2021-07-26
customer.io identify + track action

matteosdocsity:
added actions for customer.io in order to identify customers attributes and track customer events


matteosdocsity:
@davidtamaki @phillipperalez do you know what is needed to be done now in order to not wait anymore? Thanks

PR #421, opened 2021-07-28
Fetch only users or forms for slack form requests

phillipperalez:
This will change the behavior of the Slack form to by default only fetch Channels first, with an option to select users instead. The goal of this change is to reduce the amount of entries initially loaded into the form and stay within Slack API rate limits

### DIFF
--- a/src/actions/slack/legacy_slack/slack.ts
+++ b/src/actions/slack/legacy_slack/slack.ts
@@ -29,7 +29,7 @@ https://github.com/looker/actions/blob/master/src/actions/slack/legacy_slack/REA
   async form(request: Hub.ActionRequest) {
     const form = new Hub.ActionForm()
     let channelType = "channels"
-    if (request.formParams.channelType !== undefined && request.formParams.channelType !== "users") {
+    if (request.formParams.channelType !== undefined && request.formParams.channelType === "users") {
       channelType = "users"
     }
 
--- a/src/actions/slack/legacy_slack/slack.ts
+++ b/src/actions/slack/legacy_slack/slack.ts
@@ -1,6 +1,7 @@
 import * as Hub from "../../../hub"
 
 import { WebClient } from "@slack/client"
+import _ = require("lodash")
 import {displayError, getDisplayedFormFields, handleExecute} from "../utils"
 
 export class SlackAttachmentAction extends Hub.Action {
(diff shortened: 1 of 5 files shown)


src/actions/slack/legacy_slack/slack.ts
```diff
@@ -28,9 +28,13 @@ https://github.com/looker/actions/blob/master/src/actions/slack/legacy_slack/REA
 
   async form(request: Hub.ActionRequest) {
     const form = new Hub.ActionForm()
+    let channelType = "channels"
+    if (request.formParams.channelType !== undefined && request.formParams.channelType === "users") {
```
annguy3n, line 32: nit: `request.formParams.channelType !== undefined` can be `isNil` in lodash, and that'd handle `null` and `undefined` too.

src/actions/slack/legacy_slack/slack.ts
```diff
@@ -28,9 +28,13 @@ https://github.com/looker/actions/blob/master/src/actions/slack/legacy_slack/REA
 
   async form(request: Hub.ActionRequest) {
     const form = new Hub.ActionForm()
+    let channelType = "channels"
```
annguy3n, line 31: nit: this can be simplified as

```js
request.formParams.channelType !== undefined && request.formParams.channelType === "users" ? "users" : "channels"
```

wait: wouldn't it just be

```js
request.formParams.channelType || "channels"
```

annguy3n (dismissed):
just minor nits, looks good otherwise

src/actions/slack/legacy_slack/slack.ts
phillipperalez: linter didn't like the second solution

src/actions/slack/legacy_slack/slack.ts
```diff
@@ -28,9 +29,11 @@ https://github.com/looker/actions/blob/master/src/actions/slack/legacy_slack/REA
 
   async form(request: Hub.ActionRequest) {
     const form = new Hub.ActionForm()
+    const channelType = _.isNil(request.formParams.channelType) && request.formParams.channelType === "users"
```
annguy3n, line 32: it should be `||` i think, _.isNil(request.formParams.channelType) || request.formParams.channelType === users

annguy3n (approved):

PR #425, opened 2021-08-24
Facebook Ads Customer match

StevenSchrembeck:
I'm working with @vpoiesz and Zev on this one!

This action is analogous to the Google Ads Customer Match action, but for Facebook Ads.

Feedback appreciated!


vpoiesz:
@davidtamaki and I are going to review this and iterate with the consulting partner if needed. Then we'll work with you @phillipperalez for final merge approval and launch prep. We'll circle back when ready!

src/actions/facebookcustomermatch/lib/util.ts
```diff
@@ -0,0 +1,85 @@
+const countries:any = require("i18n-iso-countries")
+const usStates:any = require('us')
+const sugarDate:any = require("sugar").Date
```
davidtamaki, line 3: These packages will need to the added to package.json (`yarn add i18n-iso-countries` etc)

Unfortunately `us` does not have a license we can use in this open source repo (Apache 2.0, MIT, or BSD (2-clause or 3-clause) licenses are OK). Could we replace this with a simple lookup for state abbreviations and names?

src/actions/facebookcustomermatch/lib/executor.ts
```diff
@@ -0,0 +1,448 @@
...
+      normalizationFunction: (value: string) => getDayOfMonth(this.normalize(value))
+    },
+    {
+      lookMLFieldName: "LastName",
+      fallbackRegex: /last/i,
+      userField: "lastName",
+      normalizationFunction: this.normalize
+    },
+    {
+      lookMLFieldName: "FirstName",
```
davidtamaki, line 120: LookML objects are typically snake_case not camelcase e.g. `first_name` not `FirstName`. Can we update the lookMLFieldName values respectively? I think this will get a better hit rate

src/actions/facebookcustomermatch/lib/api.ts
```diff
@@ -0,0 +1,195 @@
...
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstName}_${formattedRow.city}_${formattedRow.state}_${formattedRow.birthYear}`, "LN_FN_CT_ST_DOBY_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstInitial}_${formattedRow.zip}`, "LN_FI_ZIP_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstInitial}_${formattedRow.city}_${formattedRow.state}`, "LN_FI_CT_ST_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstInitial}_${formattedRow.state}_${formattedRow.birthday}`, "LN_FI_ST_DOB_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstName}_${formattedRow.state}_${formattedRow.birthYear}`, "LN_FN_ST_DOBY_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstName}_${formattedRow.country}_${formattedRow.birthday}`, "LN_FN_COUNTRY_DOB_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.lastName}_${formattedRow.firstName}_${formattedRow.birthday}`, "LN_FN_DOB_SHA256"],
+    [(formattedRow: UserFields) => `${formattedRow.externalId}`, "EXTERN_ID"],
```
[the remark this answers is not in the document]
davidtamaki: Also I could very well be doing something wrong but I haven't been able to successfully submit any data with various combinations (also tried dropping the SHA256 to just `EMAIL` and listing out the various keys as a string[] as per the docs but I keep hitting: 

`debug: Facebook error message was: (#100) Data is missing or does not match schema`

In any case, I think this type of error should also be surfaced back to Looker

StevenSchrembeck:
Updated! Running a final test now to make sure my final cleanup didnt break anything

StevenSchrembeck:
note that JSONDetailLiteStream seems to always send the full results set, regardless of whether you want to limit to "in table results" or not. Better than the reverse!

src/actions/facebookcustomermatch/lib/executor.ts
```diff
@@ -0,0 +1,478 @@
...
+        winston.debug("info", `Could not match field ${columnLabel}
+        by tags or regex. Dropping it from upload.`)
+      }
+    }
+    const formattedRow = this.getFormattedRow(row, this.schema)
+    this.matchedHashCombinations = this.getMatchingHashCombinations(formattedRow,
+      validFacebookHashCombinations)
+    if (this.matchedHashCombinations.length <= 0) {
+      throw new Error("Could not match your data to any valid Facebook data combinations." +
+      "See documentation for details. No data has been sent to Facebook. CLEANFAIL")
```
davidtamaki, line 302: I tested this with some data with no columns matched. Throwing the error here makes oboe fail and causes the AH server to crash. Is there another way we can handle this?

Pasting error for reference:
```
.../node_modules/oboe/dist/oboe-node.js:2041
          throw new Error(e.message)
    at Timeout._onTimeout (.../node_modules/oboe/dist/oboe-node.js:2041:17)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
error Command failed with exit code 1.
```

StevenSchrembeck: Hmmm. The control flow is weird. We would need to kill the stream somehow then indicate to the top level `execute` that no columns were matched so it can send this 200 error.

Maybe we could pass a callback that aborts streaming because of no columns matching. Would need to see what the oboe api has to do that

StevenSchrembeck: that or try/catch inside the stream function, kill the stream, THEN throw the error. so goofy

StevenSchrembeck: okay, took another look. if you catch the error in the first function (just wrap `determineSchema(...)` in a try/catch) you can call `reject` with an error message/ or "clean fail" flag then send the response exactly like in `executor.js:203`

heck, rejecting the promise might even throw an error and then nothing needs to change

PR #438, opened 2021-10-18
fix: Use users.conversations instead of conversations.list

smaspe:
`conversations.list` contains all the available channels, and is a Tier 2 API (20 requests per minute)

`users.conversations` contain only conversations the user is already added to,  and is Tier 3 (50 requests per minute + bursts)

For organizations with large numbers of channels, this makes a huge difference, as listing all of the channels with such a limited API can take several minutes, versus under a second for the list of channels a user is already in.

Both APIs have the same signature for the intended purposes of this

See https://api.slack.com/methods/users.conversations vs https://api.slack.com/methods/conversations.list


phillipperalez (dismissed):

smaspe:
> @[smaspe](https://github.com/smaspe) smaspe dismissed [phillipperalez](https://github.com/phillipperalez)’s [stale review ](https://github.com/looker-open-source/actions/pull/438#pullrequestreview-939710683)via [dacf42a ](https://github.com/looker-open-source/actions/commit/dacf42a4df33465e2cae532db49b26489b214f67)[10 minutes ago](https://github.com/looker-open-source/actions/pull/438#event-6514524674)

Sorry about that, I had to re-author the commits so that google would recognize the CLA

phillipperalez:
I ended up porting the change over to another PR under my name (the cla thing is brand new and I really wanted to get the fix in sooner than later). As a heads up though Slack has a gotcha in the users conversation response object. We were previously filtering based on `is_member` for channels, however `is_member` doesn't exist for the user conversation responses which was an interesting find when we tried listing channels :smiley:

PR #442, opened 2021-12-23
New Action: Salesforce Campaigns

davidtamaki:
This Action allows Looker users to create campaigns in their Salesforce organization and populate the campaign members with leads and/or contacts. 

One new package has been added to manage the Salesforce API: [jsforce](https://github.com/jsforce/jsforce)

Following the same pattern as the google actions, this action also uses the AH oauth functionality and will thus require SALESFORCE_CLIENT_ID and SALESFORCE_CLIENT_SECRET to be passed as environment variables. The action will not register itself with the hub if they are not present.

Future work will be to add streaming results functionality to support larger result sets.

Looking forward to any and all feedback!


davidtamaki:
Closing this PR in favour of https://github.com/looker/actions/pull/447

PR #447, opened 2022-01-12
New Action: Salesforce Campaigns

davidtamaki:
This Action allows Looker users to create campaigns in their Salesforce organization and populate the campaign members with leads and/or contacts. 

One new package has been added to manage the Salesforce API: [jsforce](https://github.com/jsforce/jsforce)

Following the same pattern as the google actions, this action also uses the AH oauth functionality and will thus require SALESFORCE_CLIENT_ID and SALESFORCE_CLIENT_SECRET to be passed as environment variables*. The action will not register itself with the hub if they are not present.

Future work will be to add streaming results functionality to support larger result sets.

Looking forward to any and all feedback!

*UPDATE - after PR feedback, this action will also require SALESFORCE_MAX_RESULTS and SALESFORCE_CHUNK_SIZE as env variables.  Max results is max number of objects to fetch. This is used in the form builder to get existing campaigns (default is 10,000). Chunk size is the number of sObject sent per single request (limit is currently 200 records)


src/actions/salesforce/campaigns/campaigns_form_builder.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export class SalesforceCampaignsFormBuilder {
+  readonly oauthCreds: { oauthClientId: string; oauthClientSecret: string }
+
+  constructor(oauthClientId: string, oauthClientSecret: string) {
+    this.oauthCreds = { oauthClientId, oauthClientSecret }
+  }
+
+  async formBuilder(request: Hub.ActionRequest, tokens: Tokens) {
+    // const sfdcConn = await salesforceLogin(request);
```
theo-browne, line 14: remove commented out line

src/actions/salesforce/campaigns/campaigns_send_data.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.oauthCreds = { oauthClientId, oauthClientSecret }
+  }
+
+  async sendData(
+    request: Hub.ActionRequest,
+    memberIds: MemberIds[],
+    tokens: Tokens,
+  ) {
+    // const sfdcConn = await salesforceLogin(request);
+    const sfdcConn = await sfdcConnFromRequest(request, tokens, this.oauthCreds)
```
theo-browne, line 34: It looks like this block is repeated from above. Likely worth moving into its own function.

src/actions/salesforce/campaigns/campaigns_send_data.ts
```diff
@@ -0,0 +1,143 @@
...
+          throw new Error(
+            `Campaign creation error: ${JSON.stringify(newCampaign.errors)}`,
+          )
+        } else {
+          campaignId = newCampaign.id
+        }
+
+        break
+      case "append":
+        campaignId = request.formParams.campaign_name!
```
theo-browne, line 63: break

src/actions/salesforce/campaigns/campaigns_send_data.ts
```diff
@@ -0,0 +1,143 @@
...
+      column.data.map((id) => {
+        return {
+          [column.sfdcMemberType]: id,
+          CampaignId: campaignId,
+          Status: request.formParams.member_status,
+        }
+      }),
+    )
+
+    // flattan array of arrays into one big list
```
theo-browne, line 77: typo

src/actions/salesforce/campaigns/campaigns_send_data.ts
```diff
@@ -0,0 +1,143 @@
...
+      const cleanErrors = memberErrors.map((me) => {
+        const memberMessage = me.errors.errors.map((e: any) => e.message)
+        return { [me.memberId]: memberMessage }
+      })
+      const summary = `Errors with ${memberErrors.length} out of ${memberCount} members`
+      message = `${summary}. ${JSON.stringify(cleanErrors)}`
+      winston.debug(message)
+    }
+
+    return { tokens: newTokens, message }
```
theo-browne, line 128: Probably worth splitting most of these blocks into their own functions

src/actions/salesforce/campaigns/salesforce_campaigns.ts
```diff
@@ -0,0 +1,255 @@
+import * as winston from "winston"
+import * as Hub from "../../../hub"
+import { SalesforceOauthHelper } from "../common/oauth_helper"
+import { SalesforceCampaignsFormBuilder } from "./campaigns_form_builder"
+import { SalesforceCampaignsSendData } from "./campaigns_send_data"
+
+export const REDIRECT_URL = `${process.env.ACTION_HUB_BASE_URL}/actions/salesforce_campaigns/oauth_redirect`
+export const MAX_RESULTS = 10000 // how many existing campaigns to retrieve to append members to
```
theo-browne, line 8: Can we make these (MAX_RESULTS, CHUNK_SIZE) ENV variables

src/actions/salesforce/campaigns/test_salesforce_campaigns.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    it("returns an error response if there is no tokens", async () => {
+      request.attachment = { dataJSON: validLookerData }
+      const expectedResponse = makeErrorResponse({message: "Request state is missing code and redirect"})
+      const actualResponse = await action.validateAndExecute(request)
+
+      expect(actualResponse).to.deep.equal(expectedResponse)
+    })
+
+    // it("returns an error response if domain is invalid", async () => {
```
theo-browne, line 168: delete if unused

src/actions/salesforce/campaigns/salesforce_campaigns.ts
```diff
@@ -0,0 +1,255 @@
...
+  async execute(request: Hub.ActionRequest) {
+    if (!(request.attachment && request.attachment.dataJSON)) {
+      throw "No attached json."
+    }
+
+    if (!(request.formParams.campaign_name)) {
+      throw "Missing Salesforce campaign name."
+    }
+
+    const qr = request.attachment.dataJSON
```
theo-browne, line 98: Change variable name to something more descriptive

phillipperalez:
Does Salesforce use infinite tokens or long term tokens for oauth? Do they expire?

phillipperalez:
Also apologies, you got caught in the jump to Node 14

src/actions/salesforce/campaigns/test_salesforce_campaigns.ts
```diff
@@ -0,0 +1,255 @@
...
+const validLookerData = {
+    fields: { dimensions: [{ label: "Salesforce Contact Id", tags: ["sfdc_contact_id"], type: "string", name: "salesforce.contact_id" }]},
+    data: [{ "salesforce.contact_id": { value: "abc" } }, { "salesforce.contact_id": { value: "def" } }],
+}
+
+const action = new SalesforceCampaignsAction(testOAuthClientId, testOAuthClientSecret)
+
+describe(`${action.constructor.name} class`, () => {
+  // Use a sandbox to avoid interaction with other test files
+  const adsSinonSandbox = sinon.createSandbox()
```
theo-browne, line 24: Typo?

davidtamaki:
@phillipperalez The access token has a configurable validity period (default is 2 minutes), and there is an option to use a [refresh token](https://help.salesforce.com/s/articleView?id=remoteaccess_oauth_refresh_token_flow.htm&type=5&language=en_US) to request a new access token when the access token expires. The refresh tokens do not expire.

This action is set up with the refresh_token scope.

src/actions/salesforce/campaigns/campaigns_send_data.ts
davidtamaki: Turns out that listening for the on refresh event is unnecessary and I can just pass back the connection object which handles the refresh automatically. I've moved this logic to very end of both functions as the refresh event can occur at any time.

davidtamaki:
@theo-browne I've updated the PR with the your feedback comments and also for the node 14 update. Would you be able to take a final look?

theo-browne (approved):
LGTM

PR #460, opened 2022-03-02
feat: Firebase Push notification

Aatish-Molasi:


src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+	import * as admin from 'firebase-admin';
+	import * as app from 'firebase-admin/app';
+
+	export class FirebaseAction extends Hub.Action {
+
+	  name = "firebase"
+	  label = "Firebase"
+	  iconName = "airtable/airtable.png"
+	  description = "Use firebase to send push notifications to mobile."
+	  storageBucket = 'gs://looker-9d4d2.appspot.com'
```
gauravsaralMs, line 11: most probably this should be coming from some configuration

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+	  contentType = "image/jpeg"
+	  notificationOptions = {
+      priority: "high",
```
gauravsaralMs, line 14: most probably some formatting is off in this file.

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+      timeToLive: 60 * 60 * 24,
+      contentAvailable: true,
+      mutableContent: true
+	  }
+	  static firebaseAdmin: app.App
+	  params = []
+	  supportedActionTypes = [Hub.ActionType.Query]
```
gauravsaralMs, line 21: is this true ? I think this action does not support query action type

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+	  supportedFormats = [Hub.ActionFormat.JsonDetail]
+	  supportedFormattings = [Hub.ActionFormatting.Unformatted]
+	  supportedVisualizationFormattings = [Hub.ActionVisualizationFormatting.Noapply]
+
+    static setFirebaseClient() {
+      if (!FirebaseAction.firebaseAdmin) {
+        FirebaseAction.firebaseAdmin = admin.initializeApp({
+        								credential: admin.credential.cert({projectId: process.env.FIREBASE_PROJECT_ID,
+        																   clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
```
gauravsaralMs, line 30: why that replace in private key part ?

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+	  async execute(request: Hub.ActionRequest) {
+      let response = new Hub.ActionResponse({success: true})
+      try {
+        let data: any = {}
+        if (request.formParams.data) {
+          data = request.formParams.data
+          if (!data.alertId) {
+            throw "Need Valid AlertId."
```
gauravsaralMs, line 45: just checking if this should throw or this should send a 404 with this as message ?
not sure if execute handles the exception to 404 part.

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+    async uploadImage(request: Hub.ActionRequest): Promise<any> {
+      let data = request.formParams.data as any
+
+      try {
+        if (request.attachment?.dataBuffer) {
+          return new Promise((resolve, reject) => {
+            FirebaseAction.setFirebaseClient()
+            let dateString = new Date().toISOString()
+            let alertId = data.alertId ?? ""
+            const destFileName = dateString + alertId;
```
gauravsaralMs, line 138: alert id can be non unique across looker instances so may be we need something more to make this unique..

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+            const fileCloud = admin.storage().bucket(this.storageBucket).file(destFileName);
+              fileCloud.save(request.attachment?.dataBuffer!, {
```
gauravsaralMs, line 140: couple of questions 
- what is the auth mechanism for accessing this file
- for how long will this file remain in firebase server
- can there be data sovereignty issue

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+        if (params.timeToLive) {
+          this.notificationOptions.timeToLive = parseFloat(params.timeToLive)
+        }
+        if (params.priority) {
+          this.notificationOptions.priority = params.priority
+        }
+        try {
+          const userObj = JSON.parse(params.deviceIds ?? "[]") as any[]
+          let deviceIdObject = params.deviceIds as unknown as any[]
+          if (deviceIdObject.length == 0) {
```
gauravsaralMs, line 79: can we put checks like this which will lead to not sending notification before image is uploaded to firebase ?

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,167 @@
...
+            resolve()
+          }
+          await userObj.forEach(async (userDevices) => {
```
gauravsaralMs, line 82: https://stackoverflow.com/questions/51738684/await-with-array-foreach-containing-async-await

src/actions/firebase/firebase.ts
Aatish-Molasi: We need to support at least one action type
The other is dashboard and it looks like alerts does send dashboard query type.
I may have to verify this.

src/actions/firebase/firebase.ts
Aatish-Molasi: If there's no image it will throw in the earlier upload image method.
This should not be executed after

src/actions/firebase/firebase.ts
Aatish-Molasi: Will update the security document and update it here.

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,158 @@
...
+    static firebaseAdmin: app.App
+    params = []
+    supportedActionTypes = [Hub.ActionType.Query]
+    supportedFormats = [Hub.ActionFormat.JsonDetail]
+    supportedFormattings = [Hub.ActionFormatting.Unformatted]
+    supportedVisualizationFormattings = [Hub.ActionVisualizationFormatting.Noapply]
+
+    static setFirebaseClient() {
+      if (!FirebaseAction.firebaseAdmin) {
+        FirebaseAction.firebaseAdmin = admin.initializeApp({
```
gauravsaralMs, line 28: question on the way we are initialising firebaseAdmin object
- if some env variable like Private key is changed then we would need to restart action hub for it to get picked up as per current code.. is that expected and how other config changes work in action hub ?
- can the firebaseAdmin go to a `bad` state (don't know much about it) and in that case how will it get re-created.

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,158 @@
+  import * as Hub from "../../hub"
+  import * as admin from 'firebase-admin';
+  import * as app from 'firebase-admin/app';
```
gauravsaralMs, line 3:
```suggestion
  import * as firebaseAdmin from 'firebase-admin';
  import * as firebaseAdminApp from 'firebase-admin/app';
```

gauravsaralMs: ⬆️  just to improve readability

src/actions/firebase/firebase.ts
gauravsaralMs: can we instead add a new querytype enum "NONE" and use that ?
```
	  supportedActionTypes = [Hub.ActionType.None]
```

gauravsaralMs (commented):
consider using minimumSupportedLookerVersion to stop this action showing up in old looker instances...

src/actions/firebase/firebase.ts
Aatish-Molasi: @phillipperalez would this make more sense?

src/actions/firebase/firebase.ts
phillipperalez: This is expected for restarts to be neccessary on env changes. Env vars are loaded at container startup.

src/actions/firebase/firebase.ts
phillipperalez: None type would be fine and work for this case if we don't want it to show up in the scheduler modal

src/actions/firebase/firebase.ts
Aatish-Molasi: https://docs.google.com/document/d/1hTpLx7ScTV4yuW5A88iFVkjr4eGe6EUaqZ86oDC6JmU/edit#

Aatish-Molasi: Have made a doc
Will send it to the reviewers

drstrangelooker:
@Aatish-Molasi I rebased this on master and resolved some package.json references.

We are getting a build error, however...
```
yarn run v1.19.1
$ rm -rf lib; ./node_modules/.bin/tsc -d -p tsconfig.npm.json
Error: src/actions/google/common/oauth_helper.ts(3,29): error TS[23](https://github.com/looker-open-source/actions/runs/5619973692?check_suite_focus=true#step:5:23)07: Cannot find module 'googleapis/node_modules/google-auth-library' or its corresponding type declarations.
Error: src/actions/google/drive/google_drive.ts(5,41): error TS2307: Cannot find module 'googleapis/node_modules/google-auth-library' or its corresponding type declarations.
```
The directory `node_moduels/googleapis/` does not have a node_modules subdirectory even after running yarn install. `node_modules/google-auth-library/` does exist at the top level. I can make it work on my local machine by doing `cd node_modules/googleapis/ && yarn install`. I'm not sure if this is right, however.

src/actions/firebase/firebase.ts
gauravsaralMs: Just checking if this was validated?
```
The other is dashboard and it looks like alerts does send dashboard query type.
I may have to verify this.
```

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,169 @@
...
+      throw error
+    })
+  }
+
+  async uploadImage(request: Hub.ActionRequest): Promise<any> {
+    if (request.attachment?.dataBuffer) {
+      return new Promise((resolve, reject) => {
+        FirebaseAction.setFirebaseClient()
+        const destFileName = v4()
+        const fileCloud = firebaseAdmin.storage().bucket(this.storageBucket).file(destFileName)
```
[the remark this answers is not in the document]
Aatish-Molasi: https://firebase.google.com/docs/functions/quotas
We are on the blaze plan

src/actions/firebase/firebase.ts
gauravsaralMs: Me and Aatish discussed that using None would mean that helltool should also send None, but that would lead to changes in existing alert code (which means slack will also need change)
So let's not use `None`

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,170 @@
...
+
+if (process.env.FIREBASE_PROJECT_ID
+  && process.env.FIREBASE_CLIENT_EMAIL
+  && process.env.FIREBASE_PRIVATE_KEY
+  && process.env.FIREBASE_DATABASE
+  && process.env.FIREBASE_BUCKET
+  ) {
+    Hub.addAction(new FirebaseAction())
+} else {
+    winston.warn(`${LOG_PREFIX} Action not registered because required environment variables are missing.`)
```
gauravsaralMs, line 167: question: where can I see the logs from action hub in production instance

gauravsaralMs (commented):
high level looks good to me, will be good to get a 👍🏼  from SME of action hub

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,169 @@
...
+        const userObj = JSON.parse(params.deviceIds ?? "[]") as any[]
+        const deviceIdObject = params.deviceIds as unknown as any[]
+        if (deviceIdObject.length === 0) {
+          resolve()
+        }
+        for (const userDevices of userObj) {
+          const devices = userDevices as any[]
+          for (const device of devices) {
+            if (device.device_id && device.user_id) {
+              const deviceId = device.device_id.toString()
```
gauravsaralMs, line 87: is device_id correct ? or it should be deviceId because in tests it is deviceId not device**_**id

gauravsaralMs: IMO.. tests are incorrect..  device_id here is correct.

gauravsaralMs (approved):

src/actions/firebase/firebase.ts
Aatish-Molasi: device_id is what seems to be working right now

src/actions/firebase/firebase.ts
```diff
@@ -0,0 +1,163 @@
...
+        .catch( (error: any) => {
+          reject(error.message)
+        })
+    })
+    .catch((error) => {
+      throw error
+    })
+  }
+
+  async uploadImage(request: Hub.ActionRequest): Promise<any> {
```
gauravsaralMs, line 129: consider removing unused code.

phillipperalez (changes requested):

PR #485, opened 2022-06-16
Add version in firebase log

gauravsaralMs:
adding version (type of string) in the logs of firebase might help us to determine which version of action is deployed.

Also compiled some old changes are part of this PR


lib/actions/segment/segment.js
```diff
@@ -184,6 +184,10 @@ class SegmentAction extends Hub.Action {
                     if (currentObject.hasOwnProperty(key)) {
                         if (key === "value") {
                             returnVal[name] = currentObject[key];
+                            // Segment Identify Nulls #186583506
```
gauravsaralMs, line 187: most probably coming from https://github.com/looker-open-source/actions/pull/480

PR #490, opened 2022-07-21
Sanitize apostrophe's to Google Sheets overwrite

phillipperalez:
Will enable filenames such as `Somebody'sFIle.csv` to be safely passed into a query param for the Drive API


theo-browne (approved):
LGTM

PR #500, opened 2022-11-11
Jgotlieb salesforce + Jira changes

jasongotlieb:
this branch includes the changes is https://github.com/looker-open-source/actions/pull/495 (jira action updates) and also adds some updates to the Salesforce Action so that contacts and leads can be streamed to SF--allowing more updates with less API calls.


jasongotlieb:
Ran the unit tests and linter and everything is passing. 
This code updates the salesforce and jira actions. 
The salesforce action errors still flow through to the scheduled plan explore as expected. 
**The new dependency:: `adf-builder` is used to send a description to the jira ticket in the jira action that creates new tickets. This is a required custom object used by JIRA. If you don't send in their format, an error like this shows up:
`400 - {"errorMessages":[],"errors":{"description":"Operation value must be an Atlassian Document (see the Atlassian Document Format)"}}`
Here is the documentation: https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/

PR #525, opened 2023-05-16
Add slack postMessage for new upload path

phillipperalez:
Added new postMessage call for file upload V2 for Slack as it does not support markdown formatted messages

Facebook lib/ update is due to error in previous PR which did not update the lib/ directory


pfrisella (approved):
LGTM

PR #530, opened 2023-06-01
Update Slack to not use retries if the request is for an execute

phillipperalez:
Slack should not be retrying its file uploads as it is causing many duplicate sends. Only disable WebClient retries if the request is for an execute.

https://slack.dev/node-slack-sdk/web-api


lib/actions/slack/slack_client_manager.js
```diff
@@ -8,10 +8,17 @@ exports.PLACEHOLDER_WORKSPACE = "any";
...
+        return new web_api_1.WebClient(token, { retryConfig: { retries: 0 } });
+    }
+    else {
+        return new web_api_1.WebClient(token);
+    }
+};
 exports.makeSlackClient = makeSlackClient;
 class SlackClientManager {
-    constructor(request) {
+    constructor(request, isForm = false) {
```
pfrisella, line 21: Nit: Perhaps just make this explicit that this controls retries. s/isForm/disableRetries. Then in slack action, with new slack client it's obvious at that point you're making the choice to enable retries for forms.

phillipperalez: ack

theo-browne (approved):

PR #543, opened 2023-09-14
Update google ads api to v14

davidtamaki:
Migrating google ads API from v12 to v14.

API v12 will be sunset September 27, 2023 [[see timetable](https://developers.google.com/google-ads/api/docs/sunset-dates#timetable)].

API v14 will work until May 2024.


phillipperalez:
Failure unrelated

PR #563, opened 2024-01-31
Airtable OAuth2

phillipperalez:
Adds Airatable OAuth. Since the library provided by Airtable does not provide any wrappers around OAuth, we have opted to implement it on its own. Breaking with original API_KEYS sunset on Jan 31 2024

### DIFF
--- a/lib/actions/airtable/airtable.js
+++ b/lib/actions/airtable/airtable.js
@@ -11,27 +11,24 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
 Object.defineProperty(exports, "__esModule", { value: true });
 exports.AirtableAction = void 0;
 const Hub = require("../../hub");
+const crypto = require("crypto");
+const gaxios = require("gaxios");
+const qs = require("qs");
+const winston = require("winston");
 const airtable = require("airtable");
-class AirtableAction extends Hub.Action {
+class AirtableAction extends Hub.OAuthAction {
     constructor() {
         super(...arguments);
         this.name = "airtable";
         this.label = "Airtable";
         this.iconName = "airtable/airtable.png";
         this.description = "Add records to an Airtable table.";
-        this.params = [
-            {
-                description: "API key for Airtable from https://airtable.com/account.",
-                label: "Airtable API Key",
-                name: "airtable_api_key",
-                required: true,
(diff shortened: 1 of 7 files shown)


lib/actions/airtable/airtable.js
```diff
@@ -81,26 +78,190 @@ class AirtableAction extends Hub.Action {
...
+                    headers: {
+                        Authorization: `Bearer ${response.data.access_token}`,
+                    },
+                }).catch((err) => {
+                    winston.error(JSON.stringify(err));
+                    throw "Error listing bases, oauth probably bad.";
+                });
+            }
+            else {
+                return null;
```
Jonting416, line 98: Can we add logging here to make it more clear why we did not return null?  I think this would be "json state not found - OAuth missing" right?

src/actions/airtable/airtable.ts
```diff
@@ -72,26 +71,186 @@ export class AirtableAction extends Hub.Action {
...
+        url: "https://api.airtable.com/v0/meta/bases",
+        headers: {
+          Authorization: `Bearer ${(response.data as any).access_token}`,
+        },
+      }).catch((err) => {
+        winston.error(JSON.stringify(err))
+        throw "Error listing bases, oauth probably bad."
+      })
+    } else {
+      return null
```
Jonting416, line 89: We should consider logging the fact that json state is missing, likely with message "Missing json state -- OAuth probably failed"

src/actions/airtable/airtable.ts
```diff
@@ -72,26 +71,186 @@ export class AirtableAction extends Hub.Action {
...
+    const response = await gaxios.request({
+      method: "POST",
+      url: "https://www.airtable.com/oauth2/v1/token",
+      headers: {
+        "Authorization": `Basic ${encodedCreds}`,
+        "Content-Type": "application/x-www-form-urlencoded",
+      },
+      data: dataString,
+    })
+    JSON.stringify(response.data)
```
Jonting416, line 164: We aren't doing anything with this, remove?

Jonting416 (changes requested):
Overall looks good - have some nitty comments that should be resolved for code cleanliness before submission.

Jonting416 (approved):

PR #569, opened 2024-02-28
Update Airtable Flow to reset state on every request

phillipperalez:
unlike other oauth flows, Airtable uses a chained refresh_token which is reset on every /oauth2/v1/token request. Ensure state is reset in Looker to enable long lived schedules without constant manual reconsent

### DIFF
--- a/src/actions/airtable/airtable.ts
+++ b/src/actions/airtable/airtable.ts
@@ -55,7 +55,7 @@ export class AirtableAction extends Hub.OAuthAction {
       let accessToken
       if (request.params.state_json) {
         const stateJson = JSON.parse(request.params.state_json)
-        const refreshResponse = await this.refreshTokens(stateJson.params.state_json)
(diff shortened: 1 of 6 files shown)


src/actions/airtable/airtable.ts
```diff
@@ -66,37 +82,42 @@ export class AirtableAction extends Hub.OAuthAction {
...
 
   async form(request: Hub.ActionRequest) {
     const form = new Hub.ActionForm()
     try {
-      const response = await this.checkBaseList(request)
-      if (response === null) {
-        // @ts-ignore
-        throw "Error with checking baselist"
+      let accessToken
+      if (request.params.state_json) {
```
Jonting416, line 108: nit: this part of the method could be taken out of both form and execute to become `retrieveToken` or something, because this is duplicated between the two at the moment.

Jonting416 (approved):

PR #571, opened 2024-03-04
Update facebook api.ts from 16.0 to 17.0

mendozamedina:


matzeka:
Hi,

can you please update to v19?

v17 will be deprecated in May 2024 
(s. https://developers.facebook.com/docs/marketing-api/marketing-api-changelog/?locale=en_US)

PR #581, opened 2024-03-28
Adds error detail interface to ActionResponse class.

ehdunbar:
We have a project to improve error visibility for customers. This implements the new error_detail interface we are adding to the Action Response message that will then be consumed in helltool.

### DIFF
--- a/src/hub/action_response.ts
+++ b/src/hub/action_response.ts
@@ -5,13 +5,30 @@ export interface ValidationError {
   message: string
 }
 
-export class ActionResponse {
+export interface ErrorDetail {
+  /* error code associated with the error */
+  http_code?: number
+  /* A one word description of what happened */
+  reason: string
+  /* Error message that was thrown */
+  message: string
+  /* Detailed description of error written by us */
+  detail: string
+  /* one of the three services it could have erred in */
+  locationType: string
+  /* where in the service the failure occurred, which action was running when it erred */
+  location: string
+  /* url to help page listing the errors and giving detailed information about each */
+  help: string
+}
 
+export class ActionResponse {
   message?: string
   refreshQuery = false
   success = true
   validationErrors: ValidationError[] = []
   state?: ActionState
+  errorDetail?: ErrorDetail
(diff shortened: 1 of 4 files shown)


phillipperalez (approved):

src/hub/action_response.ts
```diff
@@ -5,20 +5,36 @@ export interface ValidationError {
   message: string
 }
 
-export class ActionResponse {
+export interface Errors {
```
Jonting416, line 8: nit: calling this interface with a plural name indicates that there should be multiple - this looks like it's only describing a single error.  Should be renamed.

test/test_action_response.ts
```diff
@@ -9,4 +10,24 @@ describe("ActionResponse validation", () => {
...
+    const response = new ActionResponse({success: false, errors: [errors]})
+    const jsonResponse = response.asJson()
+    chai.expect(jsonResponse.looker.errors[0]).to.equal(errors)
+  })
+
+  it("must be a valid response if errors object is not provided", () => {
+    const response = new ActionResponse({success: false})
+    const jsonResponse = response.asJson()
+    chai.expect(jsonResponse.looker).to.not.be.null
+    chai.expect(jsonResponse.looker.errors).to.be.undefined
```
Jonting416, line 31: We should validate that success is false here as well, given that we're putting it in the object definition.

Jonting416 (commented):
overall LGTM, just two small nits.

src/hub/action_response.ts
ehdunbar: I named it "Errors" because we are going to have an array of them. so `errors: Errors[]` still open to changing it to singular but I think plural works here because its an array?

Jonting416: using `errors: Error[]` is more clear in my opinion, this means it's an array of Error objects.

ehdunbar: Done

Jonting416 (approved):

PR #585, opened 2024-04-05
Replacing request library for callbacks with fetch

phillipperalez:
Adding ability to check status codes and return proper errors to Looker


Jonting416 (dismissed):

lib/hub/action_response.d.ts
```diff
@@ -3,28 +3,21 @@ export interface ValidationError {
...
 export interface Error {
-    /* error code associated with the error */
-    http_code: number
-    /* The enum version of the http_code */
-    status_code: string
-    /* Detailed description of error written by us, this should give the reason it happened and a plausible fix, and specify who is at fault */
-    message: string
-    /* where in the service the failure occurred, which action was running when it erred */
-    location: string
-    /* url to help page listing the errors and giving detailed information about each */
```
jimbert, line 16: Why did you remove these comments?

lib/hub/action_response.d.ts
```diff
@@ -3,28 +3,21 @@ export interface ValidationError {
...
+}
 export declare class ActionResponse {
     message?: string;
     refreshQuery: boolean;
     success: boolean;
     validationErrors: ValidationError[];
     state?: ActionState;
-    errors: Error[];
-    webhookId: string;
+    errors?: Error[];
```
jimbert, line 19: Is making this property optional safe?

lib/hub/action_response.d.ts
```diff
@@ -3,28 +3,21 @@ export interface ValidationError {
...
+    webhookId?: string;
```
jimbert, line 20: Is making this property optional safe?

lib/hub/action_response.js
```diff
@@ -6,7 +6,6 @@ class ActionResponse {
         this.refreshQuery = false;
         this.success = true;
         this.validationErrors = [];
-        this.errors = [];
```
jimbert, line 9: Why does this no longer need to be initialized?

src/hub/action_request.ts
```diff
@@ -180,62 +180,43 @@ export class ActionRequest {
...
-          .on("error", (err) => {
-            if (hasResolved && (err as any).code === "ECONNRESET") {
-              winston.info(`[stream] ignoring ECONNRESET that occured after streaming finished`, this.logInfo)
+        try {
+          fetch(url).then((response) => {
+            if (response.ok) {
+              response.body.pipe(stream)
+                  .on("error", (err) => {
+                    winston.error(`[stream] PassThrough stream error`, {
+                      ...this.logInfo,
```
jimbert, line 195: Why `...this` here?

drstrangelooker (approved):
LGTM

Jonting416 (approved):
LGTM but it looks like there are a lot of events that we are no longer supporting (not sure if they are in the new fetch library either) - eg: socket, abort, finish, etc.

phillipperalez:
From what I could tell node-fetch did not offer the same event granularity as the request library. You do not receive the same type of events and errors will be piped to the catch/reject rather than an event.

lib/hub/action_request.js
```diff
@@ -122,55 +122,41 @@ class ActionRequest {
...
+                                    .on("error", (err) => {
+                                    winston.error(`[stream] PassThrough stream error`, this.logInfo);
+                                    reject(err);
+                                })
+                                    .on("finish", () => {
+                                    winston.info(`[stream] PassThrough stream finished`, this.logInfo);
+                                    resolve();
+                                })
+                                    .on("close", () => {
+                                    winston.info(`[stream] PassThrough stream closed`, this.logInfo);
```
jimbert, line 142: nit: our winston logging messages all have different formats. This one is quite terse compared to the chattier messages below.

phillipperalez: I was generally trying to keep our common ones still terse, the below error messages will get passed back to looker as well as getting logged so I tried to make it more presentable

src/hub/action_request.ts
```diff
@@ -180,62 +180,41 @@ export class ActionRequest {
...
+                  .on("error", (err: any) => {
+                    winston.error(`[stream] PassThrough stream error`, this.logInfo)
+                    reject(err)
+                  })
+                  .on("finish", () => {
+                    winston.info(`[stream] PassThrough stream finished`, this.logInfo)
+                    resolve()
+                  })
+                  .on("close", () => {
+                    winston.info(`[stream] PassThrough stream closed`, this.logInfo)
```
jimbert, line 202: What should happen to the promise here?

src/hub/action_request.ts
```diff
@@ -180,62 +180,41 @@ export class ActionRequest {
...
+                  })
             } else {
-              winston.error(`[stream] request stream error`, {
-                ...this.logInfo,
-                error: err.message,
-                stack: err.stack,
-              })
-              reject(err)
+              const responseText = `There was a problem in the streaming callback Error Code: ${response.status}`
```
jimbert, line 205: 1. Instead of `Error Code`, it be more accurate to call it something like `HTTP Status`. 
1. Is there any additional info that we could log in this situation to make our lives easier?

lib/hub/action_request.js
```diff
@@ -10,8 +10,8 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
 };
 Object.defineProperty(exports, "__esModule", { value: true });
 exports.ActionRequest = exports.ActionDownloadSettings = exports.ActionVisualizationFormatting = exports.ActionFormatting = exports.ActionFormat = exports.ActionType = void 0;
+const node_fetch_1 = require("node-fetch");
```
jimbert, line 13: why is this variable named this way?

jimbert: This is out of curiosity, I see it elsewhere like on like 16 but that one makes more sense to avoid using a common word

jimbert: Oh, this is the js file, please ignore

src/hub/action_request.ts
phillipperalez: Since this is a response from looker, usually it is just a page saying you do not have authorization or 404 because the callback expired. Agreed we can change the message

package.json
```diff
@@ -42,6 +42,7 @@
     "@types/express": "^4.17.17",
     "@types/jsforce": "^1.9.33",
     "@types/node": "^18.11.11",
+    "@types/node-fetch": "^2.6.11",
```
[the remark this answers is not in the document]
phillipperalez: it was the suggested one. We can go to 3.0.3 on our version of Node

lib/hub/action_request.js
phillipperalez: js build artifact, nothing we can control

src/hub/action_request.ts
phillipperalez: good catch, close is after finish for streaming events. We'll resolve() in the close event instead

package.json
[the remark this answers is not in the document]
phillipperalez: Was an error on my part. Fixing

jimbert, line 45: This PR says it's replacing the `request` library. Can we remove the dependency?

jimbert (changes requested):

PR #588, opened 2024-05-09
refactor: Add error messaging winston logs

Itzaprado:
Adding Winston logs

Changelog: 
[ ] Add winston logs in ads_request.ts
[ ] Add logs in customer_match.ts using available function

### DIFF
--- a/src/actions/google/ads/lib/ads_request.ts
+++ b/src/actions/google/ads/lib/ads_request.ts
@@ -127,23 +127,27 @@ export class GoogleAdsActionRequest {
   async execute() {
     // 0) Do execution specific validations
     if (!this.loginCid) {
-      winston.warn(`Login account id is missing, webhookID=${this.webhookId}`)
+      winston.warn("Login account id is missing", {webhookId: this.webhookId})
       throw new MissingRequiredParamsError("Login account id is missing")
     }
     if (!["create", "append"].includes(this.createOrAppend)) {
       winston.warn(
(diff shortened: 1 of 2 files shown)


src/actions/google/ads/lib/ads_request.ts
```diff
@@ -122,17 +127,23 @@ export class GoogleAdsActionRequest {
   async execute() {
     // 0) Do execution specific validations
     if (!this.loginCid) {
+      winston.warn(`Login account id is missing, webhookID=${this.webhookId}`)
```
phillipperalez, line 130: winston.warn(`Login account id is missing`, {webhookId: this.webhookId}) is a better pattern overall, adds the information to a more standard way

phillipperalez (changes requested):

src/actions/google/ads/lib/ads_request.ts
Itzaprado: Changed

phillipperalez (approved):

src/actions/google/ads/lib/ads_request.ts
```diff
@@ -146,22 +161,26 @@ export class GoogleAdsActionRequest {
     if (this.isCreate) {
       const {newListName, newListDescription} = this.formParams
       if (!newListName) {
+        winston.warn("Name for new list is missing")
```
jimbert, line 164: should we add the webhook id here?

PR #590, opened 2024-05-09
refactor: Add error messaging winston logs to GCS

Itzaprado:
Changelog:
[ ] Add error messaging to Google Cloud Storage action

### DIFF
--- a/src/actions/google/gcs/google_cloud_storage.ts
+++ b/src/actions/google/gcs/google_cloud_storage.ts
@@ -40,7 +40,7 @@ export class GoogleCloudStorageAction extends Hub.Action {
   async execute(request: Hub.ActionRequest) {
 
     if (!request.formParams.bucket) {
-      winston.error("Need Google Cloud Storage bucket.", {webhookId: request.webhookId})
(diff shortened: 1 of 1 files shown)


src/actions/google/gcs/google_cloud_storage.ts
```diff
@@ -38,6 +40,7 @@ export class GoogleCloudStorageAction extends Hub.Action {
   async execute(request: Hub.ActionRequest) {
 
     if (!request.formParams.bucket) {
+      winston.error("Need Google Cloud Storage bucket.", {webhookId: request.webhookId})
```
jimbert, line 43: should you add the prefix here as well?

jimbert (approved):
lgtm, I love that you're adding a prefix here to make it easier to find these log messages!

In general, we ought to consider DRYing up our logging to include something like prepending the action name to the beginning of all log messages as well as including the webhook id.

src/actions/google/gcs/google_cloud_storage.ts
Itzaprado: Thanks, done.

PR #599, opened 2024-06-03
Add error info to google drive and fix dependencies

Jonting416:
This change will update the google_drive action to return additional data when errors occur during execution.  In addition, this change will fix the google-auth-library to 7.14.1 because there is a bug when updating it.

Lastly, this change also added a helper function when creating http errors.  This is a nice creation function.

### DIFF
--- a/src/actions/google/drive/google_drive.ts
+++ b/src/actions/google/drive/google_drive.ts
@@ -56,14 +56,19 @@ export class GoogleDriveAction extends Hub.OAuthAction {
         await this.sendData(filename, request, drive)
         resp.success = true
       } catch (e: any) {
-        const error: Error = errorWith(
+        let error: Error = errorWith(
             HTTP_ERROR.internal,
             "Error while sending data " + e.message,
         )
+
+        if (e.code && e.reason) {
+          error = {...error, http_code: e.code, message: `${HTTP_ERROR.internal.description} ${e.reason}`}
+        }
+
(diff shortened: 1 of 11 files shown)


src/actions/google/drive/google_drive.ts
```diff
@@ -36,18 +38,33 @@ export class GoogleDriveAction extends Hub.OAuthAction {
...
+        winston.error(`${error.message}`, {error, webhookId: request.webhookId})
         return resp
       }
       try {
         await this.sendData(filename, request, drive)
         resp.success = true
       } catch (e: any) {
+        const error: Error = errorWith(
+            HTTP_ERROR.internal,
+            "Error while sending data " + e.message,
```
ehdunbar, line 61: Do you know if google drive sends back any additional info besides `e.message`? It would be really nice if they also sent back a error code so we aren't erroring with just a 500. 

For example GCS error contains a `e.code` and `e.reason` that could be helpful for us to narrow down what the error was.

ehdunbar: https://github.com/looker-open-source/actions/blob/master/src/actions/google/gcs/google_cloud_storage.ts#L114

Jonting416: Unsure, but added a clause to handle this.

ehdunbar (approved):

PR #600, opened 2024-06-04
Refactor airtable token refresh

Jonting416:
This change will move airtable token refresh to when the attempt to reach airtable fails instead of on each request.

This aligns with the airtable style and should work better and increase availability.

### DIFF
--- a/src/actions/airtable/airtable.ts
+++ b/src/actions/airtable/airtable.ts
@@ -55,32 +55,27 @@ export class AirtableAction extends Hub.OAuthAction {
       let accessToken
       if (request.params.state_json) {
         const stateJson = JSON.parse(request.params.state_json)
-        const refreshResponse = await this.refreshTokens(stateJson.tokens.refresh_token)
-        accessToken = (refreshResponse as any).data.access_token
-        // Every single access_token invalidates previous refresh_token. Need to
-        // update state on EVERY request
-        state.data = JSON.stringify({
-          tokens: {
-            refresh_token: (refreshResponse as any).data.refresh_token,
-            access_token: accessToken,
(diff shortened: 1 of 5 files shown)


ehdunbar (dismissed):

ehdunbar:
Did you/can you manually test these changes?

ehdunbar (approved):

src/actions/airtable/airtable.ts
```diff
@@ -55,32 +55,32 @@ export class AirtableAction extends Hub.OAuthAction {
...
-        return new Promise<void>((resolve, reject) => {
-          table.create(record, (err: { message: string } | null, rec: any) => {
-            if (err) {
-              reject(err)
-            } else {
-              resolve(rec)
-            }
+      try {
+        await this.executeAirtable(request, records, accessToken)
+      } catch {
```
phillipperalez, line 69: Essentially using an exception as an if-else feels wrong in this context. Can we try to make this a flow that does not exceptions instead?

Jonting416: It's effectively more like a retry - I could rewrite it by making the logic fit inside a sub method with a retry count but since I wanted to get something done super quick I just did it like this.

The condition to trigger this is that the external call fails due to oauth failure, so then we trigger the refresh workflow.

Jonting416:
These tests have been manually verified - going ahead and rewriting the condition.

Jonting416:
After some additional discussion - noted that the current behavior does depend upon failures/exceptions to branch logic - no easy way forward right now without reconfiguring dependencies.  Can revisit upon further investment into this action.

PR #601, opened 2024-06-07
adds increased error messaging to slack

ehdunbar:


src/actions/slack/test_slack.ts
```diff
@@ -335,9 +335,20 @@ describe(`${action.constructor.name} tests`, () => {
...
+        error: {
+          documentation_url: "TODO",
+          http_code: 400,
+          location: "ActionContainer",
+          message: "Server cannot process request due to client request error. [SLACK] Missing client",
+          status_code: "BAD_REQUEST",
+        },
+        message: "Server cannot process request due to client request error. [SLACK] Missing client",
+        validationErrors: [],
+        webhookId: undefined,
```
Jonting416, line 350: We should set this in the request and make it defined.

src/actions/slack/test_utils.ts
```diff
@@ -294,8 +294,20 @@ describe(`slack/utils unit tests`, () => {
...
+                error: {
+                  documentation_url: "TODO",
+                  http_code: 400,
+                  location: "ActionContainer",
+                  message: "Server cannot process request due to client request error. [SLACK] Missing channel",
+                  status_code: "BAD_REQUEST",
+                },
+                message: "Server cannot process request due to client request error. [SLACK] Missing channel",
+                validationErrors: [],
+                webhookId: undefined,
```
Jonting416, line 309: Same here

src/actions/slack/utils.ts
```diff
@@ -1,8 +1,10 @@
-import {WebClient} from "@slack/web-api"
+import {ErrorCode, WebClient} from "@slack/web-api"
 import * as gaxios from "gaxios"
 import * as winston from "winston"
+import { HTTP_ERROR } from "../../error_types/http_errors"
 import * as Hub from "../../hub"
 import {ActionFormField} from "../../hub"
+import { Error} from "../../hub/action_response"
```
Jonting416, line 7: nit: missing space.  Does this not get caught by tslint?

src/actions/slack/utils.ts
```diff
@@ -62,6 +64,37 @@ const usableDMs = async (slack: WebClient): Promise<Channel[]> => {
     return users.map((user: any) => ({id: user.id, label: `@${user.name}`}))
 }
 
+export const setErrorResponse = (error: any) => {
```
Jonting416, line 67: I added a function here which does something similar, except you define the HttpError and message to go to the function.  Can we re-use that?

https://github.com/looker-open-source/actions/blob/master/src/hub/action_response.ts#L22

src/actions/slack/legacy_slack/slack.ts
```diff
@@ -1,7 +1,7 @@
 import * as Hub from "../../../hub"
 
 import {WebClient} from "@slack/web-api"
-import {displayError, getDisplayedFormFields, handleExecute} from "../utils"
+import {displayError, getDisplayedFormFields, handleExecute } from "../utils"
```
Jonting416, line 4: inconsistent spacing

src/actions/slack/slack.ts
```diff
@@ -36,10 +38,24 @@ export class SlackAction extends Hub.DelegateOAuthAction {
   executeInOwnProcess = true
 
   async execute(request: Hub.ActionRequest) {
+    const resp = new Hub.ActionResponse()
     const clientManager = new SlackClientManager(request, true)
     const selectedClient = clientManager.getSelectedClient()
     if (!selectedClient) {
-      return new Hub.ActionResponse({success: false, message: AUTH_MESSAGE})
+      const error: Error = {
```
Jonting416, line 45: Use errorWith

src/actions/slack/test_utils.ts
Jonting416: Meant to note to set the webhookID in this test as well

src/actions/slack/test_utils.ts
```diff
@@ -439,7 +452,7 @@ describe(`slack/utils unit tests`, () => {
             })
 
             const finalizeFailSpy = sinon.spy(async () => Promise.reject({
-                type: "CHANNEL_NOT_FOUND",
+                type: "missing_channel",
```
Jonting416, line 455: Not sure I see the purpose of this change

src/actions/slack/utils.ts
```diff
@@ -124,15 +126,27 @@ export const getDisplayedFormFields = async (slack: WebClient, channelType: stri
 }
 
 export const handleExecute = async (request: Hub.ActionRequest, slack: WebClient): Promise<Hub.ActionResponse> => {
+    const response = new Hub.ActionResponse({success: true})
     if (!request.formParams.channel) {
-        winston.error(`${LOG_PREFIX} Missing channel`, {webhookId: request.webhookId})
-        throw "Missing channel."
+        const error: Error = {
```
Jonting416, line 131: Use errorWith

src/actions/slack/utils.ts
```diff
@@ -219,8 +233,45 @@ export const handleExecute = async (request: Hub.ActionRequest, slack: WebClient
             })
         }
     } catch (e: any) {
-        winston.error(`${LOG_PREFIX} Slack App Error: ${e.message}`)
-        response = new Hub.ActionResponse({success: false, message: e.message})
+        let error: Error = errorWith(
```
Jonting416, line 236: This assignment here should be the else on the if/else clause below to avoid creating two error objects here.

ehdunbar: How would it create two error objects? Wouldn't it just override the variable error if any of the conditions were met?

Jonting416: We create an error object when we call errorWith - in this code as it is set up, we are creating a baseline object with HTTP_ERROR.internal.  Depending on the code afterwards, we may potentially replace it with another object.  This is a small update, but I think makes the code easier to represent if we just put this object into the else of the if/else clause below.

ehdunbar: The issue is that TS doesn't know `let error` var exists later in the code when we set the `response`.

Jonting416: discussed offline - might have been mixing up languages but this is fine as is.  If we take another pass let's put this into a private function

Jonting416 (approved):

PR #602, opened 2024-06-07
adds error message utility to help format errors

ehdunbar:
For GCS errors `e.code` can be either a number such as 403, or a status string such as `ECONNRESET`. This adds a utility function that maps the status string or status code to the proper error definition we have created.

Example of logs that will be fixed with this change: 

```
{ 
error: Internal server error. [Google Cloud Storage] undefined, 
http_code: 500, 
status_code: 403, 
message: Internal server error.
}
```

```
{
error: Internal server error. [Google Cloud Storage] undefined,
http_code: 500, 
status_code: ECONNRESET, 
message: Internal server error.
}
```
https://pantheon.corp.google.com/logs/query;query=SEARCH%2528%22gcs%22%2529;storageScope=storage,projects%2Factionhubpp-production%2Flocations%2Fglobal%2Fbuckets%2F_Default%2Fviews%2F_AllLogs;cursorTimestamp=2024-06-06T07:00:09.782322Z;duration=P7D?project=actionhubpp-production&e=13803378&jsmode=O&mods=monitoring_api_prod

### DIFF
--- a/src/actions/debug/debug.ts
+++ b/src/actions/debug/debug.ts
@@ -33,7 +33,8 @@ export class DebugAction extends Hub.Action {
     }
 
     return new Hub.ActionResponse({
-      message: `Completed debug action successfully by doing ${activities.join(", ")}.`, success: true
+      message: `Completed debug action successfully by doing ${activities.join(", ")}.`,
+      success: true,
     })
   }
 
--- a/src/actions/debug/debug.ts
+++ b/src/actions/debug/debug.ts
@@ -3,7 +3,9 @@ import * as winston from "winston"
 import * as Hub from "../../hub"
 
 export function registerDebugAction() {
-  Hub.addAction(new DebugAction())
+  if (process.env.ACTION_HUB_DEBUG_ENDPOINT) {
+    Hub.addAction(new DebugAction())
+  }
 }
 
 export class DebugAction extends Hub.Action {
@@ -12,7 +14,7 @@ export class DebugAction extends Hub.Action {
   label = "Debug"
   description = "Sends data to a sample website and optionally sleeps."
   supportedActionTypes = [Hub.ActionType.Cell, Hub.ActionType.Dashboard, Hub.ActionType.Query]
-  supportedFormats = [Hub.ActionFormat.JsonDetail]
+  supportedFormats = []
   params = []
   executeInOwnProcess = false
 
@@ -32,10 +34,7 @@ export class DebugAction extends Hub.Action {
       await this.delay(sleep)
     }
 
-    return new Hub.ActionResponse({
-      message: `Completed debug action successfully by doing ${activities.join(", ")}.`,
-      success: true,
-    })
+    return new Hub.ActionResponse({message: `Completed debug action successfully by doing ${activities.join(", ")}.`})
   }
 
   async form() {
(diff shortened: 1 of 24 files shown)


src/error_types/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  // Default to 500 internal server error
+  let formattedErrorInfo: HttpErrorInfo = {
+    status: HTTP_ERROR.internal.status,
+    code: HTTP_ERROR.internal.code,
+    description: `${HTTP_ERROR.internal.description} Error Format util`,
+  }
+
+  // if input is a number we want to find the corresponding error code and set
+  // the values based on that code
+  if (!isNaN(+code)) { // if e.code is not not a number; if e.code is a number
```
Jonting416, line 15: Just use isNumber instead?

src/error_types/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+    Object.values(HTTP_ERROR).forEach((errorType) => {
+      // if code is a number, set default code in case we haven't defined that specific code
+      formattedErrorInfo = {...formattedErrorInfo, code: +code}
```
Jonting416, line 18: With this logic - the code should always be set to the last touched errorType that we find in the forEach loop called on HTTP_ERROR values.  I've already implemented something similar here: https://github.com/looker-open-source/actions/blob/master/src/hub/action_response.ts#L22

src/error_types/utils.ts
ehdunbar: I was looking it up and it looks like `!isNaN` is the syntactical way to check in ts

src/actions/google/gcs/test_google_cloud_storage.ts
```diff
@@ -96,6 +96,35 @@ describe(`${action.constructor.name} unit tests`, () => {
...
+      request.attachment = {dataBuffer: Buffer.from("1,2,3,4", "utf8")}
+      return chai.expect(action.validateAndExecute(request)).to.eventually
+        .deep.equal({
+          refreshQuery: false,
+          success: false,
+          error: {
+            documentation_url: "TODO",
+            http_code: 500,
+            location: "ActionContainer",
+            message: "Internal server error. [Google Cloud Storage] Error while sending data Could not authenticate request\nerror:1E08010C:DECODER routines::unsupported",
```
Jonting416, line 119: Wait is this making an actual call out to GCS?  We should probably update this test and make it local only or else it can become flaky and depend upon the implementation of the external service.

Jonting416 (dismissed):
Will approve for now but we definitely need to tidy up these tests so they are proper unit tests.

src/actions/debug/debug.ts
```diff
@@ -3,9 +3,7 @@ import * as winston from "winston"
 import * as Hub from "../../hub"
 
 export function registerDebugAction() {
-  if (process.env.ACTION_HUB_DEBUG_ENDPOINT) {
-    Hub.addAction(new DebugAction())
-  }
+  Hub.addAction(new DebugAction())
```
Jonting416, line 6: I might have missed this in a previous commit - but do we want to update the Debug action here?  It's unrelated to the commit topic so I'm inclined to do one more commit to revert the debug changes.

Jonting416 (approved):

PR #603, opened 2024-06-08
Add error details for google_sheets action

Jonting416:
This change will add the details for the google_sheets action.

Mostly adding new tests to cover the error path.

### DIFF
--- a/lib/actions/google/drive/sheets/google_sheets.js
+++ b/lib/actions/google/drive/sheets/google_sheets.js
@@ -47,13 +47,7 @@ class GoogleSheetsAction extends google_drive_1.GoogleDriveAction {
                 const drive = yield this.driveClientFromRequest(stateJson.redirect, stateJson.tokens);
                 let filename = request.formParams.filename || request.suggestedFilename();
                 if (!filename) {
-                    const error = {
-                        http_code: http_errors_1.HTTP_ERROR.bad_request.code,
-                        status_code: http_errors_1.HTTP_ERROR.bad_request.status,
-                        message: `${http_errors_1.HTTP_ERROR.bad_request.description} ${LOG_PREFIX} Error creating file name`,
-                        location: "ActionContainer",
-                        documentation_url: "TODO",
-                    };
+                    const error = Hub.errorWith(http_errors_1.HTTP_ERROR.bad_request, `${LOG_PREFIX} Error creating file name`);
                     resp.error = error;
                     resp.success = false;
                     resp.message = error.message;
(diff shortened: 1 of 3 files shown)


lib/actions/google/drive/sheets/google_sheets.js
```diff
@@ -66,9 +76,15 @@ class GoogleSheetsAction extends google_drive_1.GoogleDriveAction {
                 }
                 catch (e) {
                     this.sanitizeGaxiosError(e);
-                    winston.error(`Failed execute for Google Sheets. Error: ${e.toString()}`, { webhookId: request.webhookId });
+                    let error = Hub.errorWith(http_errors_1.HTTP_ERROR.internal, `Failed execute for Google Sheets. Error: ${e.toString()}`);
```
ehdunbar, line 79: We've been creating `LOG_PREFIX` const instead of writing 'Google Sheets'

https://github.com/looker-open-source/actions/pull/601/files#diff-edbe3e2324dae5d23d86815669f573a8cfc8904a00b55c783ddc5ba583d9638eL17

Jonting416: Added in the new commit

src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -40,9 +42,18 @@ export class GoogleSheetsAction extends GoogleDriveAction {
 
             let filename = request.formParams.filename || request.suggestedFilename()
             if (!filename) {
+                const error: Hub.Error = {
```
ehdunbar, line 45: why don't we use `errorWith` here as well?

src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -58,10 +69,20 @@ export class GoogleSheetsAction extends GoogleDriveAction {
...
                 this.sanitizeGaxiosError(e)
-                winston.error(`Failed execute for Google Sheets. Error: ${e.toString()}`,
-                    {webhookId: request.webhookId})
+                let error: Hub.Error = Hub.errorWith(
+                    HTTP_ERROR.internal,
+                    ` ${LOG_PREFIX} Failed execute. Error: ${e.toString()}`,
+                )
+
+                if (e.code && e.errors && e.errors[0] && e.errors[0].message) {
+                    error = {...error, http_code: e.code, message: `${HTTP_ERROR.internal.description} ${e.errors[0].message}`}
```
ehdunbar, line 78: We need to have the `log_prefix` appended to the error message here as well, otherwise we won't be able to tell which action is erroring with just one log line.

src/actions/google/drive/sheets/google_sheets.ts
Jonting416: Updated

ehdunbar (approved):

PR #610, opened 2024-06-28
Fixing Quota Retries for WriteBatch requests

phillipperalez:
* Making retry backoff configurable with a default of 3 seconds (old hardcode)
* Fixing Write Quota retries to include 500 with "Quota exceeded" as a valid error to retry
* Fix read retry to not DDOS on continual 429.


src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -14,7 +14,9 @@ const MAX_REQUEST_BATCH = process.env.GOOGLE_SHEETS_WRITE_BATCH ? Number(process
 const MAX_ROW_BUFFER_INCREASE = 6000
 const SHEETS_MAX_CELL_LIMIT = 5000000
 const MAX_RETRY_COUNT = 5
+const RETRY_BASE_DELAY = process.env.GOOGLE_SHEETS_BASE_DELAY ? Number(process.env.GOOGLE_SHEETS_WRITE_BATCH) : 3
```
Jonting416, line 17: Should this number be GOOGLE_SHEETS_BASE_DELAY as well

src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -161,7 +163,7 @@ export class GoogleSheetsAction extends GoogleDriveAction {
         const spreadsheetId = files.data.files[0].id!
 
         const sheets = await this.retriableSpreadsheetGet(spreadsheetId, sheet,
-            0, request.webhookId!).catch((e: any) => {
+            1, request.webhookId!).catch((e: any) => {
```
Jonting416, line 166: magic number - can we put this under a variable so we have context on what it means?  (I assume it's retry count here??)

Jonting416: also are we just starting at retry 1 now?

phillipperalez: It is retry count. This one is overkill to put into a const IMO

src/actions/google/drive/sheets/google_sheets.ts
phillipperalez: No. The env var is more specific in case I implement it for other actions.

jimbert: Should this be its own envar then?

src/actions/google/drive/sheets/google_sheets.ts
jimbert: I think it's worth naming this for clarity

lib/actions/google/drive/sheets/google_sheets.js
```diff
@@ -20,7 +20,9 @@ const MAX_REQUEST_BATCH = process.env.GOOGLE_SHEETS_WRITE_BATCH ? Number(process
 const MAX_ROW_BUFFER_INCREASE = 6000;
 const SHEETS_MAX_CELL_LIMIT = 5000000;
 const MAX_RETRY_COUNT = 5;
+const RETRY_BASE_DELAY = process.env.GOOGLE_SHEETS_BASE_DELAY ? Number(process.env.GOOGLE_SHEETS_WRITE_BATCH) : 3;
```
jimbert, line 23: it could be helpful to include the time unit in the constant name

jimbert (dismissed):
approved with minor nits

src/actions/google/drive/sheets/google_sheets.ts
Jonting416: I meant specifically on this line we reference two env vars, I figured they should be the same env var

Jonting416 (approved):

lib/actions/google/drive/sheets/google_sheets.js
Jonting416: +1 but can be done as a fast followup

PR #613, opened 2024-07-20
Use builtin library and add test to verify that delay is actually delaying

phillipperalez:
Previous Promise<void> didn't appear to actually be applying delays

### DIFF
--- a/lib/actions/google/drive/sheets/google_sheets.d.ts
+++ b/lib/actions/google/drive/sheets/google_sheets.d.ts
@@ -25,6 +25,6 @@ export declare class GoogleSheetsAction extends GoogleDriveAction {
     retriableFileList(drive: Drive, options: any, attempt: number, webhookId: string): Promise<any>;
     flush(buffer: sheets_v4.Schema$BatchUpdateSpreadsheetRequest, sheet: Sheet, spreadsheetId: string, webhookId: string): Promise<import("gaxios").GaxiosResponse<sheets_v4.Schema$BatchUpdateSpreadsheetResponse>>;
     flushRetry(buffer: sheets_v4.Schema$BatchUpdateSpreadsheetRequest, sheet: Sheet, spreadsheetId: string, webhookId: string): Promise<import("gaxios").GaxiosResponse<sheets_v4.Schema$BatchUpdateSpreadsheetResponse>>;
-    protected delay(time: number): Promise<void>;
+    delay(time: number): Promise<void>;
     protected sheetsClientFromRequest(redirect: string, tokens: Credentials): Promise<sheets_v4.Sheets>;
 }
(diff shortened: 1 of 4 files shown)


src/actions/google/drive/sheets/test_google_sheets.ts
```diff
@@ -538,6 +538,15 @@ describe(`${action.constructor.name} unit tests`, () => {
       })
     })
 
+    describe("delay", () => {
+      it("will block on await delay", async () => {
```
jimbert, line 542: Can you provide more detail for why this is necessary?

src/actions/google/drive/sheets/test_google_sheets.ts
```diff
@@ -538,6 +538,15 @@ describe(`${action.constructor.name} unit tests`, () => {
...
+        const start = Date.now()
+        await action.delay(100)
```
jimbert, line 544: What are the units here?

src/actions/google/drive/sheets/test_google_sheets.ts
```diff
@@ -538,6 +538,15 @@ describe(`${action.constructor.name} unit tests`, () => {
...
+        const end = Date.now()
+        chai.expect(end - start).to.be.greaterThanOrEqual(10)
```
jimbert, line 546: What are the units here?

src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -444,10 +445,8 @@ export class GoogleSheetsAction extends GoogleDriveAction {
         throw `Max retries attempted`
     }
 
-    protected async delay(time: number) {
-        await new Promise<void>((resolve) => {
-            setTimeout(resolve, time)
-        })
+    async delay(time: number) {
```
jimbert, line 448: What are the units here?

lib/actions/google/drive/sheets/google_sheets.d.ts
```diff
@@ -25,6 +25,6 @@ export declare class GoogleSheetsAction extends GoogleDriveAction {
     retriableFileList(drive: Drive, options: any, attempt: number, webhookId: string): Promise<any>;
     flush(buffer: sheets_v4.Schema$BatchUpdateSpreadsheetRequest, sheet: Sheet, spreadsheetId: string, webhookId: string): Promise<import("gaxios").GaxiosResponse<sheets_v4.Schema$BatchUpdateSpreadsheetResponse>>;
     flushRetry(buffer: sheets_v4.Schema$BatchUpdateSpreadsheetRequest, sheet: Sheet, spreadsheetId: string, webhookId: string): Promise<import("gaxios").GaxiosResponse<sheets_v4.Schema$BatchUpdateSpreadsheetResponse>>;
-    protected delay(time: number): Promise<void>;
+    delay(time: number): Promise<void>;
```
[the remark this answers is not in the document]
phillipperalez: it doesn't particularly matter and then I can actually test the function

src/actions/google/drive/sheets/test_google_sheets.ts
phillipperalez: the previous implementation, although it looked correct, was not actually delaying retries. I am seeing another area which I missed so I will upload another change to this as well

PR #614, opened 2024-07-24
fix: add relax_column_count to avoid invalid record length error

Itzaprado:
Change Log:
[ ] Add relax_column_count in the parse's options to tolerate data sets with inconsistent number of fields and avoid invalid record length error


Itzaprado:
> Hi, can you add more detail to the commit message to leave context for why we needed this change?

Done.

Jonting416 (approved):

PR #615, opened 2024-07-25
fix: add function to get folder Id

Itzaprado:
The following changes were made to fix infinite loading when the user has a high storage on their Drive.

Changelog:
[ ] Add function to get folderID from provided folder name
[ ] Change parents value to use the obtained folderID
[ ] Change dropdown for an input text
[ ] Update tests with the expected form

### DIFF
--- a/lib/actions/google/drive/google_drive.d.ts
+++ b/lib/actions/google/drive/google_drive.d.ts
@@ -29,6 +29,7 @@ export declare class GoogleDriveAction extends Hub.OAuthAction {
     }[]>;
     getMimeType(request: Hub.ActionRequest): string | undefined;
     sanitizeGaxiosError(err: any): void;
+    getFolderId(request: Hub.ActionRequest, drive: Drive): Promise<string | undefined>;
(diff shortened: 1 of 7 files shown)


Jonting416:
If I understand correctly, this change will update the form to no longer be a drop down and then require the end user to input the folder ID.  Then, on execute, we will validate this folder ID by searching the list of all folders to ensure that it exists, otherwise fail because we were unable to find the folder ID?

Itzaprado:
> If I understand correctly, this change will update the form to no longer be a drop down and then require the end user to input the folder ID. Then, on execute, we will validate this folder ID by searching the list of all folders to ensure that it exists, otherwise fail because we were unable to find the folder ID?

Yes, but instead of the folder ID it should be the folder name, then on execute we search for the folder ID and if it doesn't  exist we use the drive root to save the file.

jimbert:
How does it work if there are multiple folders of the same name?

Itzaprado:
> How does it work if there are multiple folders of the same name?

The file is saved in the first folder found

Itzaprado:
Pull request closed, fix was addressed in another way in PR #622

PR #617, opened 2024-08-06
adds increased error logging to drive form request

ehdunbar:
This adds error reporting for form requests for Google Drive. We are assuming the error will be the same shape as the execute call but in case it is not, we are adding a list of object keys to the error log as well.

### DIFF
--- a/src/actions/google/drive/google_drive.ts
+++ b/src/actions/google/drive/google_drive.ts
@@ -166,7 +166,7 @@ export class GoogleDriveAction extends Hub.OAuthAction {
           errorType,
           `${LOG_PREFIX} ${e.message}`,
         )
-        let errorObjectKeys: any
+        const errorObjectKeys: any = []
         for (const [key, _] of Object.entries(e)) {
           errorObjectKeys.push(key)
         }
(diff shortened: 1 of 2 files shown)


src/actions/google/drive/google_drive.ts
```diff
@@ -161,7 +161,21 @@ export class GoogleDriveAction extends Hub.OAuthAction {
           return form
         }
       } catch (e: any) {
-        winston.error("Can not sign in to Google", {webhookId: request.webhookId} )
+        const errorType = getHttpErrorType(e, this.name)
+        let error: Error = errorWith(
+          errorType,
+          `${LOG_PREFIX} ${e.message}`,
+        )
+        const errorObjectKeys: any = []
```
Jonting416, line 169: I don't follow why we're pulling all of the errorObjectKeys out.  Are we logging this as part of the error log for future reference of what data we could log?

ehdunbar: Yes exactly, we are assuming what the error object looks like on line 173, but if we are wrong or if there are other keys on `e` then we could add them after this gets called a few times.

Jonting416 (approved):

PR #622, opened 2024-08-20
Add new method to get folder for Sheets/Drive

phillipperalez:
By default do not fetch the folders for the Sheets/Drive actions Provide a way to only fetch on explicit ask, while providing a way to enter a url of a folder and extract the folder id from it instead. Solves several issues with large Drives not being able to select all folders.


src/actions/google/drive/google_drive.ts
```diff
@@ -105,59 +106,78 @@ export class GoogleDriveAction extends Hub.OAuthAction {
...
+              required: true,
+              type: "select",
+            })
+            form.state = new Hub.ActionState()
+            form.state.data = JSON.stringify({tokens: stateJson.tokens, redirect: stateJson.redirect})
+          // We did not fetch the folder, offer to fetch or to enter a folderid
+          } else {
+            form.fields.push({
+              description: "Add url of directory that will receive the data.",
+              label: "Folder ID",
```
ehdunbar, line 161: Can we change this to `Folder URL` I can see the case where the user doesn't read the description and sees only `Folder ID` and it would fail because we can't handle just the folder id.

ehdunbar (commented):
Can we write some basic tests for this?

src/actions/google/drive/google_drive.ts
```diff
@@ -105,59 +106,78 @@ export class GoogleDriveAction extends Hub.OAuthAction {
...
+              default: folders[0].name,
+              required: true,
+              type: "select",
+            })
+            form.state = new Hub.ActionState()
+            form.state.data = JSON.stringify({tokens: stateJson.tokens, redirect: stateJson.redirect})
+          // We did not fetch the folder, offer to fetch or to enter a folderid
+          } else {
+            form.fields.push({
+              description: "Add url of directory that will receive the data.",
```
jimbert, line 160: Can we provide more explanation to the user about what this looks like? Perhaps an example of the URL

jimbert: Maybe something like: 

Enter the full Google Drive URL of the folder where you want to save your data. It should look something like https://drive.google.com/corp/drive/folders/xyz. If this is inaccessible, your data will be saved to the root folder of your Google Drive. You do not need to enter a URL if you have already chosen a folder in the dropdown menu.

src/actions/google/drive/sheets/google_sheets.ts
```diff
@@ -7,7 +7,7 @@ import { drive_v3, google, sheets_v4 } from "googleapis"
 import { GaxiosPromise } from "googleapis-common"
 import * as winston from "winston"
 import { getHttpErrorType } from "../../../../error_types/utils"
-import { Error, errorWith } from "../../../../hub/action_response"
+import { Error, errorWith } from "../../../../hub"
```
Jonting416, line 10: Did we move a dependency?

Jonting416 (approved):

jimbert:
lgtm

PR #624, opened 2024-08-29
Phillip/verify take two

phillipperalez:
Adds new scope to oauth to check email identity, then checks against a domain allowlist if it exists. If the allowlist does not exist, perform execution and forms as normal


src/actions/google/drive/google_drive.ts
```diff
@@ -368,6 +400,22 @@ export class GoogleDriveAction extends Hub.OAuthAction {
     return google.drive({version: "v3", auth: client})
   }
 
+  protected async checkDomain(redirect: string, tokens: Credentials, domainList: string) {
+    const list = domainList.split(",")
```
jimbert, line 404: Should we think about a more robust way to parse the string? One that can handle extra whitespace and other input issues?

`google.com,@googlea.com, gmail.com  ,,  , `

lib/actions/google/drive/google_drive.js
```diff
@@ -44,6 +50,19 @@ class GoogleDriveAction extends Hub.OAuthAction {
             }
             const stateJson = JSON.parse(request.params.state_json);
             if (stateJson.tokens && stateJson.redirect) {
+                if (request.params.domain_allowlist) {
```
jimbert, line 53: a whitespace allowlist string would still trigger this, should we have a more robust condition before entering this?

phillipperalez:
Run `yarn lint --fix` and then commit again

ehdunbar (approved):

PR #636, opened 2024-12-17
Fix sheets resize race condition

bgraves-lo:
* We need to wait for retriableResize to complete before attempting to paste our new data.

### DIFF
--- a/lib/actions/google/analytics/lib/csv_header_transform_stream.d.ts
+++ b/lib/actions/google/analytics/lib/csv_header_transform_stream.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="node" />
 /// <reference types="node" />
+/// <reference types="node" />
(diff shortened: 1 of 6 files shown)


lib/actions/google/analytics/lib/csv_header_transform_stream.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="node" />
 /// <reference types="node" />
+/// <reference types="node" />
```
bgraves-lo, line 3: Seems like cruft being added by `yarn build`?

phillipperalez: should be fine

drstrangelooker (dismissed):
LGTM

phillipperalez (approved):
LGTM, letting Jim see this too

PR #638, opened 2025-01-09
Update deploy docs to use Cloud Run

phillipperalez:
Use Cloud Run build file for example of self hosted action hub.


jimbert (approved):
lgtm

PR #639, opened 2025-01-16
Update Firebase from deprecated library to newer supported library

phillipperalez:
This includes required Node and Typescript updates. JS compile target is now at ES2018


src/actions/firebase/firebase.ts
```diff
@@ -96,12 +96,13 @@ export class FirebaseAction extends Hub.Action {
...
               const deviceId = device.device_id.toString()
               notificationData.user_id = device.user_id.toString()
-              const payload = {
+              const message = {
+                token: deviceId,
                 notification,
                 data: notificationData,
               }
               try {
-                await this.sendMessageToDevice(deviceId, webhookId, payload, notificationOptions)
```
bgraves-lo, line 104: Not blocking, but does `notificationOptions` even need to exist if we're not passing it anywhere?  Looks like we're using it above to modify `this.notificationOptions`, but seems like you could just get rid of the const and directly modify `this.notificationOptions`?

src/actions/firebase/firebase.ts
```diff
@@ -121,14 +122,12 @@ export class FirebaseAction extends Hub.Action {
...
+  async sendMessageToDevice(message: firebaseAdmin.messaging.TokenMessage,
                             webhookId: string | undefined,
-                            payload: firebaseAdmin.messaging.MessagingPayload,
-                            options?: firebaseAdmin.messaging.MessagingOptions,
-                            ): Promise<any> {
+  ): Promise<any> {
     return new Promise<void>((resolve, reject) => {
       FirebaseAction.setFirebaseClient()
-      firebaseAdmin.messaging().sendToDevice(deviceId, payload, options)
+      firebaseAdmin.messaging().send(message)
```
bgraves-lo, line 130: Does `send` no longer need options?  Just trying to understand the purpose of `this.notificationOptions`

bgraves-lo: From Phillip:
> looks like those notification options don't exist anymore

bgraves-lo (approved):
LGTM

PR #642, opened 2025-01-22
Additional log and error message for Sheets

phillipperalez:
Adding an additional log to clarify the error in the case of no credentials being present for google sheets. 
Also adding an error message to be populated back to Looker.


bgraves-lo (approved):
LGTM

PR #650, opened 2025-02-21
Sync Streaming

phillipperalez:
Use the readable event instead of data to guarantee no race conditions for uploading data to google sheets and resizing


theo-browne (approved):
LGTM

PR #655, opened 2025-03-04
Slack migration from v1 upload

phillipperalez:
~This is currently a breaking change requiring `channel:manage`, `im:write`,`mpim:write`, and `groups:write` added to the scope permissions~

Managed to rewrite in a way that requires no scope creep. If you pass a UXXXX token to `chat.postMessage()` you will receive an implicit conversion to direct message token types (DXXXXX). This was not documented by slack documentation.

### DIFF
--- a/src/actions/slack/utils.ts
+++ b/src/actions/slack/utils.ts
@@ -124,7 +124,16 @@ export const getDisplayedFormFields = async (slack: WebClient, channelType: stri
 
     return response
 }
-
+const convertUMTokens = async (channel: string, slack: WebClient): Promise<string> => {
+    const openResponse = await slack.conversations.open({channel: channel})
+    if (openResponse.error) {
+        throw openResponse.error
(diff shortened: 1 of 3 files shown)


jimbert (approved):

src/actions/slack/utils.ts
```diff
@@ -167,40 +166,59 @@ export const handleExecute = async (request: Hub.ActionRequest, slack: WebClient
                             {webhookId},
                         )
 
-                        // Unfortunately UploadV2 does not provide a way to upload files
```
Jonting416, line 170: Is this comment now defunct?  Do we have a way to upload files to user tokens with UploadV2 now?

Jonting416 (approved):

src/actions/slack/utils.ts
phillipperalez: We don't, we have a way to convert the tokens though

PR #659, opened 2025-03-12
Add string as possibility for query_id

phillipperalez:
Looker is moving to using strings instead of numbers for query_ids. Ensure we support this,


aschott-looker (approved):
LGTM!

PR #661, opened 2025-03-27
chore!: updating azure-storage package to new @azue/storage-blob package

Itzaprado:
BREAKING CHANGE: azure-storage package is now deprecated


src/actions/azure/azure_storage.ts
```diff
@@ -85,18 +76,22 @@ export class AzureStorageAction extends Hub.Action {
...
+      }
     })
   }
 
-  private azureClientFromRequest(request: Hub.ActionRequest) {
+  private azureClientFromRequest(request: Hub.ActionRequest): BlobServiceClient {
     try {
-      return azure.createBlobService(request.params.account!, request.params.accessKey!)
+      const sharedKeyCredential = new StorageSharedKeyCredential(request.params.account!, request.params.accessKey!)
+      return new BlobServiceClient(
```
Jonting416, line 91: I see that we're moving away from using Azure storage to Azure storage-blob.  This is a different service from a cursory search.

Is there no factory method similar to `azure.createBlobService` that we can depend on, or is there no special initialization logic that is being used?

Itzaprado: No, since the azure-storage package is now deprecated and as per their migration guide, this is the best way to do it. 

https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/storage/storage-blob/MigrationGuide.md#constructing-the-clients

phillipperalez (approved):
LGTM

PR #665, opened 2025-04-15
[Hubspot] Migrate to use Private App Access Token

tomsseisums:
As per https://developers.hubspot.com/docs/guides/apps/private-apps/migrate-an-api-key-integration-to-a-private-app, HubSpot expects to work with Private App Access Token's now.

The change in this PR makes the HubSpot actions work with said Private App Access Token instead.

Otherwise Hubspot actions are unchanged.

### DIFF
--- a/src/actions/hubspot/hubspot.ts
+++ b/src/actions/hubspot/hubspot.ts
@@ -44,9 +44,9 @@ export class HubspotAction extends Hub.Action {
   iconName = "hubspot/hubspot.png"
   params = [
     {
-      description: "An api key for Hubspot.",
-      label: "Hubspot API Key",
-      name: "hubspot_api_key",
+      description: "Access Token of your Private Hubspot App.",
+      label: "Hubspot Access Token",
+      name: "hubspot_access_token",
       required: true,
       sensitive: true,
     },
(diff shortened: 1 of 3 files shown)


phillipperalez (approved):

tomsseisums:
@phillipperalez when can I expect this to be released to production and available in Looker's default Action Hub?

phillipperalez:
I expect to get this out next week

tomsseisums:
I see you've released this, at least on GitHub, but I don't see the Action having it's params updated in the Action hub on my Looker instance, where it still asks for the old API key instead of Access Token.

Is that expected?

ryanmwetzel:
> I see you've released this, at least on GitHub, but I don't see the Action having it's params updated in the Action hub on my Looker instance, where it still asks for the old API key instead of Access Token.
> 
> Is that expected?

I've been banging my head off the keyboard trying to figure out why it's not working and this seems to be the reason. Getting error `async delivery failed due to errors "Required setting \"Hubspot Access Token\" not specified in action settings."` because it's still listed as API key in the action settings...my best guess.

ryanmwetzel:
> I expect to get this out next week

@phillipperalez any chance you can check on the above to see if that's causing an issue?

PR #668, opened 2025-04-25
Enforce Max delivery size on Slack (Slack's own limit)

phillipperalez:
Slack has a 1 GB limit for files which would fail anyways, and the buffer would occasionally overflow memory. Send back an informative error instead.

hubspot.js file was just missed in a previous merge, it is code that should have already been included


jeffrey-martinez (approved):
Adheres to slack's limitation. LGTM.

PR #673, opened 2025-05-19
Adds checks to ensure error message gets correctly extracted.

ehdunbar:
There have been customer issues where they are receiving an error 'error on child: {}'. This usually indicates that the error has non-enumerable properties. This change adds a check to see if the error is of type Error. It also adds increased logging to tell us if the error cannot be Stringified.

### DIFF
--- a/src/xpc/execute_process.ts
+++ b/src/xpc/execute_process.ts
@@ -31,7 +31,8 @@ process.on("message", (req) => {
             } else {
                 errorString = String(err)
             }
-            winston.error("Error on child: " + errorString)
+            const request = Hub.ActionRequest.fromIPC(req)
+            winston.error(`Error on child: ${errorString}. WebhookID: ${request.webhookId}`)
             process.send!({success: false, message: errorString})
         })
 })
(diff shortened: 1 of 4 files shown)


src/xpc/execute_process.ts
```diff
@@ -19,8 +19,19 @@ process.on("message", (req) => {
...
+            } else if (typeof err === "object" && err !== null) {
+                try {
+                    errorString = JSON.stringify(err)
+                } catch (jsonError: any) {
+                    errorString = `[Object could not be stringified: ${jsonError.message || jsonError.toString()}]`
+                }
+            } else {
+                errorString = String(err)
+            }
+            winston.error("Error on child: " + errorString)
```
phillipperalez, line 34: Please add the the webhookId

src/xpc/extended_execute_process.ts
```diff
@@ -22,8 +22,19 @@ process.on("message", (req) => {
...
+      } else if (typeof err === "object" && err !== null) {
+          try {
+              errorString = JSON.stringify(err)
+          } catch (jsonError: any) {
+              errorString = `[Object could not be stringified: ${jsonError.message || jsonError.toString()}]`
+          }
+      } else {
+          errorString = String(err)
+      }
+      winston.error("Error on child: " + errorString)
```
phillipperalez, line 37: same here, webhookId

phillipperalez (approved):

PR #678, opened 2025-06-04
Refactor: replace deprecated 'binary' encoding with Buffer

c3retc3:
As per the Node.js documentation, 'binary' encoding is outdated and can lead to unexpected behavior or data loss.

1. Replaced deprecated 'binary' string encoding with direct Buffer operations for data key handling.
2. The encrypt method now passes the dataKey buffer directly to crypto.update.
3. The decrypt method now correctly reconstructs the dataKey using Buffer.concat.

This improves code quality and future compatibility without altering the cryptographic logic.


c3retc3 (commented):
Explicitly declaring Buffer[] to prevent accidental pushes or incorrect types during maintenance

Issue #137, opened 2018-05-08
Tray integration lacks look metadata


gustavoatexporo:
Currently the Tray integration does not support sending look metadata (e.g., look id, look name, url, etc), only the actual look results are posted/transferred.

PR #734, opened 2026-04-23
Handle "reset" state and allow Google Ads oauthCheck to return false

jeffrey-martinez:
- Added a guard clause in oauthExtractTokensFromStateJson to handle the literal string 'reset' sent by Looker to clear state, avoiding noisy JSON parse errors.
- Replaced hardcoded return true in GoogleAdsCustomerMatch.oauthCheck with a truthful check that verifies tokens in state_json, breaking the infinite login loop for fresh logins.
- Added unit tests for both changes.
- Added clarifying comments explaining the why behind these fixes.


src/actions/airtable/airtable.ts
jeffrey-martinez, line 1: I was getting confusing JSON parse errors in the UI when I was resetting the Airtable oauth login flow. We were trying to parse the actual string "reset" as JSON. This change allows a more graceful recognition of the reset logic.

src/actions/google/ads/customer_match.ts
jeffrey-martinez, line 1: Previously we had hardcoded oauthCheck as true, during the bug bash we discovered this can result in the Oauth flow stuck looping back to the "Sign in with Google" button. This change actually verifies the connection, freeing the user from the loop once the set up completes. 

Also adds more granular error classification (line #137)

src/hub/oauth_action.ts
jeffrey-martinez, line 1: This is part of the fix for noisy JSON errors when we passed the string literal "reset" mentioned in airtable.ts.

lib/actions/airtable/airtable.js
```diff
@@ -120,12 +160,16 @@ class AirtableAction extends Hub.OAuthAction {
                 await this.checkBaseList(accessToken);
                 if (form.state === undefined) {
                     form.state = new hub_1.ActionState();
-                    if (tokens) {
+                    if (isEncrypted && tokens) {
```
rogermichoud, line 163: Does that change the logic somehow? Before if tokens was null we would not do the form.state.data = request.params.state_json.

now if tokens is null, we have a different behavior than beore

lib/actions/google/ads/customer_match.js
```diff
@@ -75,6 +82,7 @@ class GoogleAdsCustomerMatch extends Hub.OAuthAction {
         }
     }
     async form(hubReq) {
+        var _a, _b, _c, _d;
```
rogermichoud, line 85: Can you give a bit more context on why we are using 4 alphabetical variables? why are they 4? why are they non-descript?

lib/actions/google/ads/customer_match.js
```diff
@@ -107,11 +115,15 @@ class GoogleAdsCustomerMatch extends Hub.OAuthAction {
                 return wrappedResp.returnError(err);
             }
             // Other errors from the API client - typically an auth problem
-            if (err.code) {
+            const isAuthError = ((_a = err.response) === null || _a === void 0 ? void 0 : _a.status) === 401 ||
```
rogermichoud, line 118: This is scary. I don't understand what the _a, _b, etc are standing for, and why we are doing this cryptic aphabetical error code comparison. There must be a clearer way to write this, whatever this is trying to acheive.

rogermichoud (commented):
Few comments to address around clarity and logic preservation.

src/server/server.ts
jeffrey-martinez, line 1: Oauth_check previously delivered empty JSON payloads to the browser client and left a vulnerability to allow unauthenticated remote actors. This changes allows only verified requests.

lib/actions/google/ads/customer_match.js
jeffrey-martinez: This does not exist in the actual ts code, this is just what the compiled .js generates.

lib/actions/google/ads/customer_match.js
jeffrey-martinez: Same answer as above: This does not exist in the actual ts code, this is just what the compiled .js generates.

We can filter the PR for only .ts files and also ignore anything not in the src/ directory
https://github.com/looker-open-source/actions/pull/734/changes?file-filters%5B%5D=.ts

lib/actions/airtable/airtable.js
jeffrey-martinez: I think the result is unchanged. If tokens are missing or null we would throw an error at line 158 above, and would never proceed down to the if/else block that includes line 171.

ehdunbar (approved):

mnvsngv (approved):

PR #748, opened 2026-05-26
Security hardening for Google OAuth actions

mnvsngv:


src/server/server.ts
```diff
@@ -148,11 +149,35 @@ export default class Server implements Hub.RouteBuilder {
...
+        return
+      }
+
       if (isOauthAction(action) || isOauthActionV2(action)) {
-        const parts = uparse.parse(req.url, true)
-        const state = parts.query.state
-        const url = await (action as OAuthAction).oauthUrl(this.oauthRedirectUrl(action), state)
-        res.redirect(url)
+        const actionId = req.params.actionId
+        const isTargetAction = ["google_drive", "google_sheets", "mock_oauth_action"].includes(actionId)
```
jeffrey-martinez: Ideally we should not add any hardcoded action name logic inside of server.ts, we want the server to be agnostic.

Can we add a boolean to oauth_action.ts 

Gemini recommends:
    +  /**                                                                                                                                  
    +   * Indicates whether this OAuth action utilizes double-submit cookie/nonce                                                           
    +   * validation for enhanced CSRF protection.                                                                                          
    +   */                                                                                                                                  
    +  usesCsrfProtection = false                                                                                                           
    +    

That way individual actions can opt-in. LEt me know if you need more clarification

src/server/server.ts
```diff
@@ -148,11 +149,35 @@ export default class Server implements Hub.RouteBuilder {
...
+
+        winston.info(`Oauth route hit for actionId: ${actionId}, isTargetAction: ${isTargetAction}`)
+
+        if (isTargetAction) {
+          const nonce = crypto.randomBytes(16).toString("hex")
+          res.cookie("action_hub_state", nonce, {
+            httpOnly: true,
+            secure: true,
```
jeffrey-martinez: We need to consider if this `secure:true` will affect staging or local tests. 

We could try `secure: req.secure || process.env.NODE_ENV === "production",`

Did this work fine with local testing?

src/server/server.ts
```diff
@@ -183,13 +208,51 @@ export default class Server implements Hub.RouteBuilder {
...
+            throw new Error("Invalid state parameter format.")
+          }
+
+          const nonce = parts[0]
+          originalState = parts.slice(1).join(".")
+
+          const cookies = this.parseCookies(req.headers.cookie)
+          const expectedNonce = cookies.action_hub_state
+
+          if (!expectedNonce || expectedNonce !== nonce) {
```
jeffrey-martinez, line 233: Let's log this error, 

```
if (!expectedNonce || expectedNonce !== nonce) {                                                                                        
 +   this.logError(req, res, `CSRF validation failed for action ${actionId}: Nonce mismatch.`);                                            
      res.status(403).send("CSRF validation failed.");                                                                                      
      return;                                                                                                                               
    }      
```

src/actions/marketo/test_marketo.ts
```diff
@@ -212,19 +212,23 @@ import { MarketoTransaction } from "./marketo_transaction"
...
+            },
+            campaign: {
+              request: spy.campaignRequest,
+            },
+            list: {
+              addLeadsToList: spy.listAddLeadsToList,
+              removeLeadsFromList: spy.listRemoveLeadsFromList,
+            },
+          }
+        })
```
jeffrey-martinez, line 231: Should we have a matching after block for this?

```
after(() => {
       sinon.restore()
})
```

jeffrey-martinez (changes requested):
Left some comments, let me know if I can clarify

src/server/server.ts
mnvsngv: Thanks Jeff, that's a good point. I've tweaked this & I'll upload a revised or new commit soon.

src/actions/marketo/test_marketo.ts
mnvsngv: Yes we do, thanks for catching it!

src/server/server.ts
mnvsngv: Good idea.

src/server/server.ts
mnvsngv: Great idea, thanks!

I actually didn't test it locally with `secure: true` - I added it and directly tested it in the staging environment. Let me test this out once more, this time locally as well as on staging, and if everything looks good I'll upload a new commit.

mnvsngv (commented):
Thanks for the feedback, I've incorporated all of it & updated the PR.

jeffrey-martinez (approved):

